Issue #29, opened 2020-04-19
Add Contributor documentation


robdrosenberg:
Add these documents to docusarus and set up buttons to link to them on the home page.
- [x] Contributing guide
- [x] Code of Conduct 
- [x] Manifesto

flybayer:
Fyi, I just updated the manifesto in the main repo

PR #35, opened 2020-04-20
Task / Contribution Pages

coreybrown89:
- adds code of conduct page
- adds manifesto page
- adds how to contribute page
- updates edit this page url to blitz.com github repo

#29


.alexignore
```diff
@@ -0,0 +1,2 @@
+docs/introduction/manifesto.md
+docs/contributing/code-of-conduct.md
```
flybayer, line 2: What is alex complaining about? We can update the docs if needed.

coreybrown89: It blew up for the code of conduct because of the mentions of sex / gender and stuff and for some reason alex did not like our use of the word _easy_ in the manifesto.

flybayer: Ok haha, then this should be fine.

PR #40, opened 2020-04-22
Document the RPC specification (#21)

### LINKED ISSUE #21: Document the RPC specification
We need to officially document the RPC specification. Most users won't need to know about this, but it is needed for advanced use cases.

Specification is described here: https://github.com/blitz-js/blitz/issues/85

MrLeebo:
Screen Shot 2020-04-22 at 5 11 03 PM

Creates some initial documentation for the RPC specification as requested at https://github.com/blitz-js/blitz/issues/85

It's pretty barebones and copied almost verbatim from the slack comments currently, but I expect we'll have more opportunities to polish up the documentation as the design solidifies.

Resolves #21

### DIFF
--- a/docs/advanced/rpc-specification.md
+++ b/docs/advanced/rpc-specification.md
@@ -1,0 +1,104 @@
+---
+id: rpc-specification
+title: RPC Specification
+sidebar_label: RPC Specification
+---
+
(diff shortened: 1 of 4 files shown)


robdrosenberg:
@MrLeebo Thank you for adding these docs!

PR #42, opened 2020-04-27
add video player component

robdrosenberg:
This is a simple responsive video player component using the [react-player](https://github.com/CookPete/react-player) library.

I'd like to replace the CSS file with what ever solution we decide on for styling for the site.

I ripped the @flybayer alpha walkthrough off of twitter and just threw it on the home page.

We'll definitely want to improve upon this but should work for a start. I was thinking it would be cool to add some progression features to it if we have video tutorials / courses!

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,8 @@
     "@docusaurus/preset-classic": "2.0.0-alpha.50",
     "classnames": "^2.2.6",
     "react": "^16.8.4",
-    "react-dom": "^16.8.4"
+    "react-dom": "^16.8.4",
+    "react-player": "^2.0.0"
   },
   "resolutions": {
     "babel-plugin-dynamic-import-node": "2.3.0"
--- a/src/components/VideoPlayer/index.js
+++ b/src/components/VideoPlayer/index.js
@@ -1,0 +1,21 @@
+import React from "react";
+import ReactPlayer from "react-player";
+import "./video-player.css";
+
+const VideoPlayer = ({ url }) => {
+  {
+    return (
+      <div className="player-wrapper">
+        <ReactPlayer
+          className="react-player"
+          url={url}
+          width="100%"
+          height="100%"
+          controls={true}
+        />
+      </div>
+    );
+  }
+};
+
+export default VideoPlayer;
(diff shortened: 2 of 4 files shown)


coreybrown89:
@robdrosenberg bumping docusaurus to the latest alpha seems to fix the build issue locally.

```
    "@docusaurus/core": "2.0.0-alpha.54",
    "@docusaurus/preset-classic": "2.0.0-alpha.54",
```

robdrosenberg:
@coreybrown89 Looks like upgrading to v.54 didn't take 😔
Am I missing something here?

Followed their instructions on updating manually, also tried using yarn upgrade locally
https://v2.docusaurus.io/docs/installation

Are you able to build locally?

coreybrown89:
> Are you able to build locally?

@robdrosenberg hmm - I was able to build locally with the new version. 

Here's where its failing in the github actions: https://reactjs.org/docs/error-decoder.html/?invariant=294

```
ReactDOMServer does not yet support Suspense.
```

Wondering if we need to use experimental versions of `react` and `react-dom` like blitz

robdrosenberg:
@coreybrown89 I'm getting the same error locally.

robdrosenberg:
Alright, was able to get some help and figured it out! 

On build, docusaurus was trying to render the video component on the server. To fix it we had to make sure the video component only renders client side.

Docusaurus docs outlines two ways to accomplish that. 
Using [ExecutionEnvironment](https://v2.docusaurus.io/docs/next/docusaurus-core/#executionenvironment) or [BrowserOnly](https://v2.docusaurus.io/docs/next/docusaurus-core#browseronly).

I opted to use ExecutionEnvironment.

Here's some semi pseudo code for future reference:
```
import ExecutionEnvironment from '@docusaurus/ExecutionEnvironment';
render:
  if (ExecutionEnvironment.canUseDOM) {
    return YourComponent
  }
  return null
```

mgm1313:
Nice work! I think `<VideoPlayer url={useBaseUrl("/static/video/alpha_walkthrough.mp4")} />` should be `<VideoPlayer url={useBaseUrl("/video/alpha_walkthrough.mp4")} />` [without the 'static' part](https://docusaurus.io/docs/en/api-pages#using-static-assets). 
It's working as is on local development, but not on the deployment preview on Vercel. 

Furthermore I guess it would be nice to have a `maxHeight` on the video, so it can't be taller then the viewport. Could work on that if you want?

robdrosenberg:
@mgm1313 🤦‍♂️ whoops lol. Would be happy to have your help on it. I think adding maxHeight is a good idea! I'll go ahead and change the url right now.

PR #45, opened 2020-04-29
Make sure the video height doesn't exceed the viewports height.

mgm1313:
As discussed with @robdrosenberg in https://github.com/blitz-js/blitzjs.com/pull/42#issuecomment-620845293 . 
My solution reads a bit hacky 🤓 so if somebody has suggestions to make it cleaner or better readable i'd love to hear!

### DIFF
--- a/src/components/VideoPlayer/video-player.css
+++ b/src/components/VideoPlayer/video-player.css
@@ -1,6 +1,8 @@
 .player-wrapper {
   position: relative;
-  padding-top: 62.5%; /* Player ratio: 100 / (1152 / 720) */
+  margin: 0 auto;
+  padding-top: min(calc(100vh - 60px), calc(720 / 1152 * 100%)); /* minimum value of: (viewport height - header), (aspect ratio height of video) */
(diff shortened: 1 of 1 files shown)


robdrosenberg:
Hi @mgm1313 thanks for submitting this, the player scales a lot better now! My only concern is hard coding a value based on the header height. I don't see the header height changing that much but it might after the design / branding comes through. Do you think there is a way we can avoid having to update this value anytime there's a change?

mgm1313:
Yeah I've been thinking about this @robdrosenberg, and get your concern. We could use a static max-height of `85vh` instead of the current `100vh - 60px`. Although that would fail on really tight viewports (height wise). I did some tests and `85vh` seems the best trade off for the ~90% use-case.  
Unfortunately there's no method I know of with css to retrieve the actual height of the header and use that as a variable.

robdrosenberg:
@mgm1313 There might be a method once we introduce our decided on CSS solution! For now it's great. I'll merge and we can visit this again when we convert the component.

PR #46, opened 2020-04-29
Add more vertical space to features section.

mgm1313:
Ok I had anticipated on making some more quick wins 🤣 , but those were all inside theme components. Didn't want to touch on them since they're not stable yet. 

So this has been a nice introduction into Docusaurus and Infima, and as an outcome I think the added vertical space gives some more 'breathing space'  to the features section. 

Let me know what you think @merelinguist

### DIFF
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -81,7 +81,7 @@ function Home() {
       </header>
       <main>
         {features && features.length && (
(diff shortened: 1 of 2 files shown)


merelinguist:
Looks good, thank you! 🏅

PR #57, opened 2020-05-12
Add Theme UI

robdrosenberg:
Took a first stab at it but couldn't get it to override the current styles with the time I had.

Layout Component is in the @theme directory which is default docusaurus code, not sure how to add the ThemeProvider component from theme ui to it to start overriding ALL the default styles.

I'll push up the yarn package and an initial theme.js file. The settings in the theme.js aren't final but just some initial theme values to work with.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,8 @@
     "classnames": "^2.2.6",
     "react": "^16.8.4",
     "react-dom": "^16.8.4",
-    "react-player": "^2.0.0"
+    "react-player": "^2.0.0",
+    "theme-ui": "^0.3.1"
   },
   "resolutions": {
     "babel-plugin-dynamic-import-node": "2.3.0"
--- a/src/theme/theme.js
+++ b/src/theme/theme.js
@@ -1,0 +1,35 @@
+export default {
+  fonts: {
+    body:
+      'Roboto, -apple-system, BlinkMacSystemFont, "Segoe UI", "Helvetica Neue", sans-serif',
+    heading: '"Libre-franklin", serif',
+    monospace: "Menlo, monospace",
+  },
+  fontSizes: [12, 14, 16, 20, 24, 32, 48, 64],
+  fontWeights: {
+    body: 400,
+    heading: 700,
+    bold: 700,
+  },
+  lineHeights: {
+    body: 1.5,
+    heading: 1.125,
+  },
+  letterSpacings: {
+    body: "normal",
+    caps: "0.2em",
+  },
+  colors: {
+    text: "#F4F4F4",
+    background: "#45009D",
+    primary: "##6700EB",
+    secondary: "#45009D",
+  },
+  styles: {
+    root: {
+      // uses the theme values provided above
+      fontFamily: "body",
+      fontWeight: "body",
+    },
+  },
+};
--- a/yarn.lock
+++ b/yarn.lock
@@ -161,7 +161,7 @@
   dependencies:
     "@babel/types" "^7.8.3"
 
-"@babel/helper-module-imports@^7.8.3":
+"@babel/helper-module-imports@^7.0.0", "@babel/helper-module-imports@^7.8.3":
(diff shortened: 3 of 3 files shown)


alexfinnarn:
Adding the ThemeProvider like in https://theme-ui.com/getting-started has always worked for me. I usually make a Layout component and add it there.

I'm just a lurker who probably won't help too much so take this with a grain of salt...but people seem to like Theme UI + Gatsby a lot and using MDX as well. Since the current docs are in Markdown, it shouldn't be too hard to convert them to MDX and Theme UI has specific features for styling in MDX. Just a thought if Docasaurus proves troublesome. I don't think you'd have much trouble with Gatsby + Theme UI for making a documentation site.

robdrosenberg:
@alexfinnarn we'll take any help we can get! I appreciate the extra info. We're using Docusaurus v2 for the Blitz site. I don't think we'd be able to switch over to gatsby and fulfill the requirements we have for the site. I use gatsby for development at work and I'm a big fan of it!

As for adding the theme provider, a Layout component already exists in the docusaurus theme. I'm not sure how or if we can override it while keeping the site stable as docusaurus updates it's alpha versions.

We'll definitely want to utilize theme ui's MDX styling because the docusaurus v2 docs are MDX!

robdrosenberg:
@jasonblalock Would you have any ideas on how to solve this with the existing Layout component from Docusaurus?

robdrosenberg:
@jasonblalock Yea, we might be out of luck for the time being with v2 Docusaurus and overwriting the theme.

@flybayer Feel free to drop your reasons, but I think we wanted to go with Docusaurus v2 because it had a lot of nifty docs features baked in. Algolia, doc versioning, ready for translations. It's also created by Facebook.
Currently it seems v2's customizability is just not where we need it at the moment, and we knew going into this v2 might cause some issues since it's in Alpha.

Maybe switching over to gatsby is the best bet for now. I'm not sure, I'd hate to throw away the work that's already been done. It'd certainly take some of the headaches away but would probably take more development time in the future. I'm sure there's some gatsby starters / recipes that could help though.

flybayer:
Well shucks.. We'll just have to build a lot more stuff from scratch if we don't use docusuarus. But then it is a big benefit to have full control over everything. That's probably what we should do..

I think we should definitely use Next instead of Gatsby. That way we have more flexibility to do more dynamic stuff in the future if we want. And it'll also be more familiar to Blitz users/contributors than Gatsby.

flybayer:
Wow yeah, Chakra looks nice! Chakra and Rebass both use styled-system but not theme-ui explicitly (theme-ui is built on styled-system). But that's fine since Chakra has build in theming.

coreybrown89:
Whoops, my mistake. I agree though, Chakra looks promising. Thanks to @robdrosenberg for showing it to me.

coreybrown89:
Btw - didn't mean to close this. That was an accidental click

flybayer:
I want those of you who want to build out the site to choose the css solution among these top choices.

alexfinnarn:
A note on Rebass and theme-ui...I thought Rebass is an ancestor to theme-ui created by the same author years ago. 

https://www.npmtrends.com/rebass-vs-theme-ui-vs-@chakra-ui/core

If you go out to the 2 year timeline, you'll see the trend than made me refactor from starting with Rebass and going to theme-ui. I didn't have to change much, `<Box>` needed the `sx` prop for styles in place of using the prop styles, e.g. `<Box sx={{p:4}}>` vs. `<Box p={4}>`. But everything else was the same.

I don't know much about Chakra. It looks nice. The only downside appears to be bundle size, although you might be able to shake, shake, shake it off if only using parts. I do wish some of the Chakra components were in theme-ui, but then again, that would increase the bundle size...idk, I'm know expert on bundling code, just my two cents.

flybayer:
🎉 Theme UI now installed via #63!

PR #58, opened 2020-05-12
Update default favicon

dwightwatson:
This replaces the default favicon (which I believe comes from the HTML 5 boilerpalte) with the purple Blitz logo.

image

image

This was generated using the logo from the art repo.


flybayer:
Thank you @dwightwatson!

PR #59, opened 2020-05-19
first pass at formalizing CLI docs

aem:
Now that the CLI has stabilized a bit, I wanted to sit down and write up some formal docs on the CLI. There's some room for improvement here, and the `console` isn't documented very well, but it'll be good enough for a solid introduction.


docs/cli/commands.md
```diff
@@ -4,61 +4,209 @@ title: CLI Commands
...
-Generate a new blitz project at `<current_folder>./NAME`
+### `blitz help`
+
+Displays all of the CLI's information in the console. Additionally, all core commands support a `--help` flag which will print the information in this document to the console. It's omitted from all "Options" tables below for brevity.
+
+### `blitz new`
+
+Generates a new Blitz project in the current directory. The project will include basic scaffolding for a landing page, a database, and all of the other configuration that makes Blitz's magic work.
+
+#### Arguments
```
aem, line 19: Every command now shares the same structure and a similar writing style, so all of the sections feel cohesive

docs/cli/commands.md
```diff
@@ -4,61 +4,209 @@ title: CLI Commands
...
+| Argument    | Shorthand | Description                                                                                                     | Default |
+| ----------- | --------- | --------------------------------------------------------------------------------------------------------------- | ------- |
+| `--dry-run` | `-d`      | Displays what files would be generated but does not write the files to disk.                                    | `false` |
+| `--js`      |           | Generates a JavaScript-only project. If omitted, a TypeScript project will be generated by default.             | `false` |
+| `--npm`     |           | Uses `npm` as the project's package manager. If omitted, the project will default to `yarn` if it is installed. | `false` |
+
+#### Examples
+
+```bash
+blitz new task-manager
```
aem, line 40: All examples are from an actual app rather than using contrived foobar examples

src/css/custom.css
```diff
@@ -22,3 +22,11 @@
   margin: 0 calc(-1 * var(--ifm-pre-padding));
   padding: 0 var(--ifm-pre-padding);
 }
+
+.prism-code.language-bash
```
aem, line 26: This allows the `Copy` button on code sections to work without copying the leading character. The first code line in bash scripts will display the caret `:before` the element, so when the user selects the text in the code panel the `>` **won't be selectable.**

image

PR #65, opened 2020-05-25
Fix relative links carried over from docs migration

pixelmord:
1. QUestion: Remove the link to the contributing guide, currently there is not more in the blitzjs repo then here.
2. linked the new issue page
3. fixed all relative links that 404ed after the migration


docs/contributing/future-of-blitz.mdx
```diff
@@ -22,9 +22,9 @@ Blitz is only getting started, and it's going to take an entire community to bri
 How you can help:
 
 1. Tell others about Blitz
-2. Report bugs by opening an issue here on GitHub
+2. Report bugs or request features by opening an issue on [Blitz' GitHub repo](https://github.com/blitz-js/blitz/issues/new/choose)
 3. Send us feedback in the [Blitz Slack](https://slack.blitzjs.com).
-4. Contribute code. We have a lot of issues that are ready to work on! Start by reading [The Contributing Guide](contributing/how-to-contribute.md). Let us know if you need help.
```
robdrosenberg, line 27: We'll want this link to remain here so let's see if we can fix it.
It looks like the link might be broken for a couple reasons. I have ideas to fix it but I don't have time to double check it tonight. Feel free to give it a try if you have time! It may take both options to fix it.

1. I believe @flybayer made all the documents .mdx files maybe the link isn't working because it's trying to link to a .md file?
2. [Here](https://v2.docusaurus.io/docs/markdown-features#referencing-other-documents) is a link to the docusaurus docs on referencing other documents. Maybe we need to tweak how the document is referenced? Could try changing it to the following:
`[The Contributing Guide](./how-to-contribute.mdx)`
`[The Contributing Guide](how-to-contribute.mdx)`

pixelmord: I noticed the same issue on other pages, the reason behind this is most likely that those docs were copied from the GH blitzjs as MD files and therefore the links are still the same as in the other repo.
It should be an easy fix and I can go through all the files and fix linking issues.

However, the one mentioned Sentece here in 4. was intentionally removed for another reason: Since the docusaurus brings a navigation sidebar, you do not need to link to another page in every case. In this case you are already reading one sub-section of the contribution guide, so I thought why have a link here to a parent page on the same website...

So my proposal:
1. Let me fix all the links, revert the removal of 4. for nowand add this to this PR
2. In the meantime you or @flybayer or .. can quickly give a statement on the planned linking between the docs/website and the project-GH that might also have info for contributors. And then the result needs either another issue/pr or can also be quickly incorporated in this PR

flybayer: I think lots of internal linking is great. For this one, yes, the Code of Conduct is there in the sidebar, but it's still much easier to have it linked right in the text :)

pixelmord: OK I will change these links (contributing, code of conduct) then to all stay on the blitz.com page

flybayer: Thanks! :)

flybayer: Once this goes live, we'll remove those from the main repo.

flybayer:
@pixelmord is this all ready to go now?

PR #78, opened 2020-06-12
Fix some broken links

exclipy:


exclipy:
I scanned the site with deadlinkchecker.com and the only two remaining are the "you can see here" links at https://blitzjs.com/docs/api-routes

They look like they should link to https://nextjs.org/docs/api-routes/api-middlewares and https://nextjs.org/docs/api-routes/response-helpers respectively. Should I link there or were those pages supposed to be copy/pasted in to blitzjs.com?

flybayer:
@exclipy awesome, thank you for this!!

And yes, those links on the api routes page should link to the proper page on blitzjs.com

exclipy:
My question was that I couldn't find the proper page on blitzjs.com,so should I make them or should I link to nextjs.org?

flybayer:
@exclipy ah yeah!

I just realized they are actually on the same page.

"Response Helpers" and "Default Middlewares"

exclipy:
Ah of course, how did I miss that

flybayer:
Perfect, thank you! :)

PR #83, opened 2020-06-15
Update docs to show current generator and console usage

wKovacs64:
We'll want to wait for the next Blitz release before merging this (depends on https://github.com/blitz-js/blitz/pull/659), but it should be ready to review at least.

I started out with the goal of updating the REPL examples to use top-level await, but ran into overall flow issues with the tutorial now that model generation is in (specifically, manually creating the model definitions in `db/schema.prisma` followed by `blitz generate all` yields undesired results). Per feedback in Slack from Brandon and Adam, I made some fairly significant updates to the tutorial to reflect the current recommended way of doing things (as I perceive them, anyway). As a result, this turned into a bigger update than I had originally intended, but I think it results in a better user experience.

### DIFF
--- a/docs/cli-console.mdx
+++ b/docs/cli-console.mdx
@@ -16,13 +16,17 @@ This makes it very simple to read or write to your database and run one-off comm
 
 You have entered the Blitz console
 Tips: - Exit by typing .exit or pressing Ctrl-D
-      - Use your db like this: db.project.findMany().then(console.log)
-      - Use your queries/mutations like this: getProjects().then(console.log)
-      - Top level `await` support coming: https://github.com/blitz-js/blitz/issues/230
+      - Use your db like this: await db.project.findMany()
+      - Use your queries/mutations like this: await getProjects()
 ✔ Loading your code
 
-⚡ > db.question.findMany().then(console.log)
+⚡ > await db.question.findMany()
 [
-  { id: 1, text: 'What’s up?', publishedAt: 2020-04-24T22:08:17.307Z }
+  {
+    id: 1,
+    createdAt: 2020-06-15T15:06:14.959Z,
+    updatedAt: 2020-06-15T15:06:14.959Z,
+    text: "What's up?"
+  }
 ]
 ```
--- a/docs/cli-generate.mdx
+++ b/docs/cli-generate.mdx
@@ -7,7 +7,7 @@ sidebar_label: blitz generate
 
 Use this command to scaffold all the boring code into your project.
 
-Can generate pages, queries, and mutations. Generating `Prisma` models is coming soon. Also coming soon is support for custom templates based on the built-in templates so you can customize the generator to your app's needs.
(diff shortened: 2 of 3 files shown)


flybayer:
Looks good! Could you also update the contents of the generated queries/mutations while you're at it? What's in the tutorial is outdated

wKovacs64:
> Looks good! Could you also update the contents of the generated queries/mutations while you're at it? What's in the tutorial is outdated

Sure thing. Ran into some other issues while doing so that are hindering accurate documentation, so I'm converting this to draft for now.

wKovacs64:
Updated some more stuff. There were some minor inconsistencies between the docs, what gets generated (e.g. TypeScript), and the final result after after the user makes modifications and saves (e.g. Prettier formatting). I tried to meet in the middle, but suggestions/corrections are welcome of course.

For these new docs to be entirely accurate, there are a few issues/PRs that need to be closed:

- [x] https://github.com/blitz-js/blitz/pull/659
- [x] https://github.com/blitz-js/blitz/pull/695
- [ ] https://github.com/blitz-js/blitz/issues/696
- [ ] https://github.com/blitz-js/blitz/issues/698

wKovacs64:
I *think* this is ready even with those last 2 outstanding issues since one is a bug that should be fixed soon and the other is very minor. Even with those discrepancies, the docs in this PR should be more accurate than the current docs.

wKovacs64:
TIL [cascading deletes](https://www.prisma.io/docs/guides/database-workflows/cascading-deletes) are not yet implemented in Prisma 2. Deleting an isolated piece of data works but deleting relational data (e.g. a `Question` that has a `Choice`) errors:

```
The change you are trying to make would violate the required relation 'ChoiceToQuestion' between the `Choice` and `Question` models.
```

The tutorial currently includes delete mutations, which work in the early steps where the simplest example question is created with no choices (which is arguably incorrect, since choices are required, but that's another topic) but will fail by the end when the relational nature of the data is utilized.

Some possible paths forward:

1. Change the example used in the tutorial to avoid relational data.
1. Apply some nasty hacks in Blitz to work around the issue (with TODOs to remove once cascading deletes land in Prisma).
1. Pretend it works, as it will at some point. Possibly include a disclaimer that deleting is currently theoretical and will error, but assuring readers it will work eventually.

Thoughts @flybayer @aem?

flybayer:
We should definitely keep in relational data so people learn how to do that.

The workaround for no cascading deletes is to manually change the deleteParent mutation to be like this:

```
await deleteChildren({where: {parentId: parentId})
await deletePartent({where: {id: parentId})
```

wKovacs64:
Yeah agreed, definitely prefer to keep the relational data.

Guess I had tunnel vision on the deletes. 🥴 I'll update it to what you have.

flybayer:
Great, thank you so much!!

PR #84, opened 2020-06-17
Remove extra bracket in local middleware example

shaunchurch:
Tiny correction to remove an unnecessary bracket in the local HTTP middleware example.


flybayer:
Awesome, thanks Shaun!

PR #94, opened 2020-06-21
Update docs with new Prisma 2 pagination

osirvent:
Update docs according https://github.com/blitz-js/blitz/pull/658


flybayer:
Thank you @osirvent!

PR #97, opened 2020-06-22
Correct typo in exotic, suggest change to unusual

shaunchurch:
Noticed exotic was typod, but while correcting wondered if 'exotic' is the best choice of word here. It's often used by westerners to mean `foreign to westerners`. Perhaps unusual is a more inclusive choice.


merelinguist:
Oh very nice catch, thank you Shaun!

PR #100, opened 2020-06-26
Add Optimistic Updates section and details for refetch option in useQuery mutate()

shaunchurch:
Add a simple sentence explaining the option to disable refetch when calling mutate. https://github.com/blitz-js/blitz/pull/715

Added in all locations it was relevant, but if it's too much detail for the useQuery/useInfiniteQuery pages I can remove those and leave it only on the Mutation Usage details. Thoughts?

### DIFF
--- a/docs/mutation-usage.mdx
+++ b/docs/mutation-usage.mdx
@@ -89,3 +89,60 @@ export default function (props) {
 #### Automatically
 
 We want to somehow automatically invalidate your query cache. We're not sure how to do this yet, but we have [an open issue that you're welcome to contribute to!](https://github.com/blitz-js/blitz/issues/586)
+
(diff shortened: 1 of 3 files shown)


flybayer:
@shaunchurch what you have looks good! Could you also add a section to https://blitzjs.com/docs/mutation-usage specifically for **Optimistic Updates**?

shaunchurch:
@flybayer Ok, first draft in. I'm using the line highlighting from #101 , so this now depends on that being merged first.

flybayer:
Beautiful!!

flybayer:
@shaunchurch can you either update this PR or create a new one so that it's a pr into the `main` branch. `master` has been deleted.

Issue #98, opened 2020-06-22
Add syntax line highlighting


flybayer:
The default docusurus Code component [had support for highlighting specific lines](https://v2.docusaurus.io/docs/markdown-features/#line-highlighting) in a code block. But we lost that when we switched to the Prism component provided by theme-ui.

However, we really need the line highlight functionality to it's easy to highlight certain changes in the tutorial.

[Here's where it's used in our source code](https://github.com/blitz-js/blitzjs.com/blob/master/src/components/MDXComponents/index.js#L14).

I don't know if the theme-ui prism component can be extended to add this? If not, then we can take the source code of the theme-ui prism component and add that directly into your codebase. Then merge in the source code that does line highlighting from the docusaurus code component.

shaunchurch:
Looks like support for highlighting specific lines was added to Theme-UI/Prism in the coming version 4.0, currently in RC1 release.

We can upgrade @theme-ui/prism separately to the rest of theme-ui so perhaps it's not a big risk to move to the release candidate just for Prism? Will submit a PR shortly.

PR #101, opened 2020-06-26
Enable highlighting specific lines within code blocks

shaunchurch:
Closes: #98 

Upgraded to @theme-ui/prism 4.0-rc.1 for support for commented code highlighting like so:

```
const example = "not highlighted";
// highlight-start
const highlighted = true;
// highlight-end
const alsoHighlighted = true // highlight-line
```

Adds a custom CSS class to visually distinguish the highlighted lines.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ When tagging a new version, the document versioning mechanism will:
 
 [See the Docusaurus documention on versioning](https://v2.docusaurus.io/docs/versioning/#tagging-a-new-version)
 
-## Comment Highlights
+## Code Syntax Highlights
 
 Code blocks are syntax highlighted using Prism. You can call attention to specific sections in a code block using comments to begin and end a block `// highlight-start` `// highlight-end`, or highlight a single line `// highlight-line`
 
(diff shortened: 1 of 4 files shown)


shaunchurch:
The CSS has a few magic numbers. I'm not sure if there's a better way to do this with theme-ui?

There's an issue on narrow screens where the highlighted div doesn't extend beyond the initial width of the <pre> that is scrollable. I don't know if there's way to force it full width within the <pre> tag.

Another option is to enable wrapping inside the <pre> tags with this:

```
pre.prism-code {
  white-space: pre-wrap;
  word-break: break-word;
}
```

shaunchurch:
image

flybayer:
@shaunchurch nice!!! Magic numbers are fine for now. We can revisit it once the full design is done.

One last thing, can you add a section to the readme of this repo on how to add line highlighting?

shaunchurch:
Now with line numbers. "The try catch block on lines 8-12 in the example..."

They're in a :before pseudo element and done entirely in CSS so aren't included when selecting or copy and pasting examples.

image

flybayer:
Awesome!! The line numbers are great!

But is there a way to disable line numbers for `sh` and `bash` type?

image

shaunchurch:
@flybayer ah good spot. Yes, they each have a language class as well, so I've disabled sh, bash and the class that seems to apply to unspecified language back tick blocks.

flybayer:
Yay!! Thank you!!!

PR #104, opened 2020-06-29
Update default question to include choice

sbardian:
In the tutorial you update `new.tsx` to create a default question "Do you love Blitz?".  Then later make updates for "Listing Choices" in the `QuestionsList` component and `[questionsId].tsx`.  After we are given an example of what the page should look like, but no where do we add choices for this default question, which might create confusion if the choices code is working/etc.  

Adding the "Yes!" choice while creating the default question will make the result after listing choices match the tutorials outcome.


flybayer:
Awesome, thank you!!

PR #105, opened 2020-06-30
Adding a "What is Next.js?" section

garrisons:
This is based off of a conversation in slack that can be found here: https://blitzjs.slack.com/archives/CUA5KK1BR/p1593475057295400

What I'm trying to do here is provide a little context if a person is reading the docs and thinks "what is Next.js?". I feel this is enough info to get the gist and worse case scenario they can learn more at their site.

I adapted this from this doc [here](https://docs.magic.link/technologies/next-js-integration) and removed some info I felt like would only add confusion.

Also, I renamed the `Why would someone use Blitz instead of Next.js?` to `Why use Blitz instead of Next.js?` just to make it more readable.


flybayer:
This is great, thank you @garrisons!

PR #115, opened 2020-07-21
Added syntax highlighting to tutorial

### LINKED ISSUE #103: Update tutorial to use the new code syntax highlighting
We now have [code syntax line highlighting support](https://github.com/blitz-js/blitzjs.com#code-syntax-highlights). So we need to update the tutorial to use this. It'll be much easier to follow! :)

LoriKarikari:
This closes #103 

I've added syntax highlighting to the tutorial page but I ran into some small issue or maybe I overlooked something.
In the last part, I wanted to highlight the code blocks separately instead of highlighting it all as I did now. I don't know if that makes a huge difference for the user but I stumbled upon some weird behavior as you can see in the screenshots below. When I've already selected a piece of code, `handleVote()`, and I want to select the questions list it doesn't highlight the right code. It's off by two lines. When I remove the highlight at `handleVote()` it does work.

I've tried to 'cheat' to two-line offset in the second part but I wasn't able to grab the piece of code so I ended up highlighting the whole thing instead. 

highlight-code

highlight-doc

### DIFF
--- a/docs/tutorial.mdx
+++ b/docs/tutorial.mdx
@@ -279,8 +279,10 @@ export default async function deleteQuestion(
   {where}: DeleteQuestionInput,
   ctx: Record<any, any> = {},
 ) {
+  // highlight-start
   // TODO: remove once Prisma supports cascading deletes
   await db.choice.deleteMany({where: {question: {id: where.id}}})
+  // highlight-end
   const question = await db.question.delete({where})
 
   return question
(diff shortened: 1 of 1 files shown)


flybayer:
Thanks @LoriKarikari!! And yes I've noticed that faulty behavior too. Could you open an issue here on it? Hopefully someone will fix it :)

PR #121, opened 2020-07-30
Fix directory name in query definition docs

kripod:


merelinguist:
Thank you, nice spot @kripod!

PR #127, opened 2020-08-11
Set NODE_ENV to production when migrating

dwightwatson:
I noticed that when trying to deploy to Vercel it was hanging at the migration step. It looks like this issue was resolved in blitz-js/blitz#786, but depends on `NODE_ENV` being set to production for the build step. I've updated the docs to reflect the change required for the migrations to proceed.


flybayer:
Good catch, thanks @dwightwatson!

Issue #129, opened 2020-08-13
usePaginatedQuery docs example incorrect


scisteffan:
### What is the problem?

When I use the example usePaginatedQuery code in the docs (https://blitzjs.com/docs/use-paginated-query) the next button is not appearing / disappearing correctly.

flybayer:
I'm not sure if this is still an issue or not

scisteffan:
This was resolved in the pull request, closing

PR #130, opened 2020-08-13
usePaginatedQuery docs example incorrect #129

scisteffan:
I have updated the docs for usePaginatedQuery and useInfiniteQuery. Both of them now reference the projects db table (rather than products) so anyone can copy/paste the code into a new blitz project.

The usePaginatedQuery example has been updated to calculate the hasMore logic on the server, in line with bug fix #858 in the blitz repo.


flybayer:
Thanks @scisteffan! I'll go ahead and merge this and we can update this again if needed from any further changes in https://github.com/blitz-js/blitz/pull/881

PR #140, opened 2020-08-26
Added region to render template

SigurdMW:
Since the webapp and the database needs to be in the same region for the `DATABASE_URL` to work without hassle, I added it


flybayer:
Thanks @SigurdMW! Are they not both in the same region by default?

SigurdMW:
They might be but I hade this issue since I created the database manually and encountered a general connection error that was hard to debug. Turns out it was due to different regions. But feel free to close this PR if you don't agree.

flybayer:
Ok makes sense, thanks!

PR #141, opened 2020-08-28
Remove boilerplate from auth docs

dwightwatson:
This is a minor change in terms of structure, but removes all of the boilerplate that is now available standard in a Blitz app.

* Model definitions
* Middleware installation
* Auth queries, mutations, and validation
* Auth error boundaries

I think there's more we can do to make the docs easier to digest, but the first step is to remove all the heavy content that's no longer required.

### DIFF
--- a/docs/auth.mdx
+++ b/docs/auth.mdx
@@ -3,7 +3,7 @@ title: Auth
 sidebar_label: Auth
 ---
 
-Blitz handles auth for you out of the box - making it easy to manage sessions, authenticate users and authorize their actions. Blitz's session management is based upon the [SuperTokens library](https://supertokens.io/), whose CTO [Rishabh Poddar](https://twitter.com/rishpoddar) designed and oversaw the implementation. In your fresh Blitz app your `db/schema.prisma` file is prefilled with models for `User` and `Session`, and `blitz.config.js` has the session middleware configured. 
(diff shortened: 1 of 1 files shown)


docs/auth.mdx
```diff
@@ -19,57 +15,9 @@ Session management performs the following functions:
 2. Attribute multiple requests to the same user, even when they are logged out
 3. Protection against CSRF attacks
 
-#### Design & Implementation
-
-Blitz session management is based on the state of the art [SuperTokens library](https://supertokens.io/). [Rishabh Poddar](https://twitter.com/rishpoddar), the CTO of SuperTokens, designed and oversaw our implementation. We're extremely grateful for Rishabh's help! 🙏
```
flybayer, line 24: Let's leave this paragraph in here as a standalone paragraph so Rishabh's recognition doesn't get overlooked within another paragraph :)

docs/auth.mdx
```diff
@@ -627,79 +441,3 @@ export default async function getUser({where}: GetUserInput, ctx: {session?: Ses
   return user
 }
 ```
-
-## Error Handling
```
flybayer, line 631: Let's leave this Error Handling section, but update the content to reflect our latest defaults. You could paste the relevant code here from the default `_app.tsx` file.

flybayer:
Sweet, thank you so much!!

PR #143, opened 2020-08-30
Add note for zsh users running generate command with default values

ivandevp:
This note can help zsh users following the tutorial until https://github.com/blitz-js/blitz/issues/724 is fixed 😃


flybayer:
Thanks Ivan!!

PR #145, opened 2020-08-31
Remove outdated auth caveat

yhoiseth:
Auth is now built-in, so there is no need to have “coming soon” any more :)


flybayer:
Awesome, thank you!

PR #148, opened 2020-09-03
Update Vercel deployment: DO transaction mode

janpio:
Prisma works fine in the default "Transaction" mode.


LoriKarikari (approved):
Thanks @janpio!

PR #156, opened 2020-09-09
#966 Adds documentation for secureProxy option to passportjs.mdx

PixelsCommander:


docs/passportjs.mdx
```diff
@@ -24,6 +24,7 @@ export default passportAuth({
   strategies: [
     // Provide initialized passport strategy here
   ],
+  secureProxy: true,
```
flybayer, line 27: Do you think this should be set to true by default?

PixelsCommander: Documentation was never my strong skill. By default, it is undefined and middleware not applied. By adding this to the code I just wanted to show that this option is there and it is boolean. Please suggest what would be a better wording/presentation for this.

flybayer (commented):
Ah bummer. I forget to actually hit submit on these yesterday.

I just published a new blitz stable release, so I'll go ahead and make these changes.

PR #157, opened 2020-09-10
Updated passportjs.mdx

sirmyron:
Removed single quotes from `redirectUrl` example and added an additional example using pathname from router.


flybayer:
Perfect, thank you!!

PR #163, opened 2020-09-16
Add step when switching from SQLite to PostgreSQL

taylorcjohnson:
Helps close issue [#1044](https://github.com/blitz-js/blitz/issues/1044) in the `blitz-js/blitz` repo.

The db provider value is no longer an array with `sqlite` and `postgres`. When switching from using SQLite to PostgreSQL the developer will need to specify the new db provider value in `db/schema.prisma`


flybayer (approved):

LoriKarikari:
Thanks for the PR @taylorcjohnson!

PR #164, opened 2020-09-16
Reorder what-is-nextjs and why-blitz pages

tsriram:
I think it'd be easier for readers to follow the docs when we have "What's Next.js" before "Why Blitz instead of Next.js?" :)


flybayer:
Good idea, thanks!

PR #173, opened 2020-09-25
Explanation how to setup local PostgreSQL database with docker-compose.

engelkes-finstreet:
A short explanation how to setup a PostgreSQL database with docker-cmpose for an easier transition to PostgreSQL.

### DIFF
--- a/docs/database-overview.mdx
+++ b/docs/database-overview.mdx
@@ -49,3 +49,46 @@ datasource db {
 
 2. In `.env.local`, change `DATABASE_URL`. For convenience, there is a commented-out PostgreSQL `DATABASE_URL` there already. Depending on your setup, you may need to modify the URL.
 3. Run `blitz db migrate`. This command will create the database (if it does not already exist) and tables based on your schema.
+
+### Run PostgreSQL database with docker-compose
+
+The easiest way to get a PostgreSQL database on your machine is to run PostgreSQL inside a docker container.
+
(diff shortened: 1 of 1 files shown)


docs/database-overview.mdx
```diff
@@ -49,3 +49,46 @@ datasource db {
...
+
+2. Inside your `.env` file add 3 new environemnt variables which are required by docker
+
+```text
+POSTGRES_USER=your_user
+POSTGRES_PASSWORD=your_password
+POSTGRES_DB=your_database_name
+```
+
+3. Change the `DATABASE_URL` variable as described in the step above
```
flybayer, line 84: I think we can remove this step here since it's already include above

docs/database-overview.mdx
```diff
@@ -49,3 +49,46 @@ datasource db {
...
+volumes:
+  data:
+```
+
+2. Inside your `.env` file add 3 new environemnt variables which are required by docker
+
+```text
+POSTGRES_USER=your_user
+POSTGRES_PASSWORD=your_password
+POSTGRES_DB=your_database_name
```
flybayer, line 81: Can you add an example here of what these three should be based on the full DATABASE_URL? (and use the DATABASE_URL that's generated in new apps)

flybayer (changes requested):
Sweet, thanks for doing this @engelkes-finstreet!

Got some suggested changes.

PR #176, opened 2020-09-25
Update deploy-vercel.mdx

garrisons:
I deployed my BlitzJS application and noticed this was missing from the docs.


flybayer:
Thanks!

PR #210, opened 2020-10-02
Test recipe locally

engelkes-finstreet:
I knew that I saw the explanation how to do this in Slack but I think it would be nice to have this inside the docs as well.


flybayer:
Thanks!!

PR #219, opened 2020-10-09
Updates authorization docs to match new Ctx

ntgussoni:
This PR matches the DOCs with the new Ctx type after 0.24.0


LoriKarikari:
Thanks @ntgussoni!

LoriKarikari (approved):

PR #229, opened 2020-10-18
Make tutorial consistent with generated code

zanedb:
While building my first Blitz project, I noticed a few inconsistencies between Blitz's generated code & the tutorial, which I aim to fix here.

I've never used Blitz before, so I apologize if any changes I made are incorrect.

Thanks! :)


flybayer:
Looks good, thank you!!

PR #230, opened 2020-10-19
fix: docker.compose.yml to docker-compose.yml

goleary:
default path is: `./docker-compose.yml`, not `./docker.compose.yml`

https://docs.docker.com/compose/compose-file/#service-configuration-reference


flybayer:
Hey, thank you!! Sorry for the delay!

PR #231, opened 2020-10-19
Fix example for connect middleware

nksaraf:
The example for the connect middleware uses the `cors` handler which actually exports a constructor. So it needs to be called before being used.


flybayer:
Thank you!! And sorry for the delay! :)

PR #232, opened 2020-10-21
Fix typo on "how the community operates" text

vitordino:
it looks like a really nice project (: anxious to test it out..

just started sketching something with next + prisma, and this might fit better


flybayer:
Thank you!

PR #233, opened 2020-10-27
Update deploy docs on how to have multiple providers in Prisma

victorferreira:


theZieger:
I'm not a maintainer but I would prefer a normal code block over a diff for this documentation. Copy and paste is easier with normal code blocks. When using a diff you have to get rid of the + signs manually. But that's just my opinion. :slightly_smiling_face: 
Anyways, good to see this beeing added to the documentation. :+1:

victorferreira:
I actually agree with you, but decided to keep the pattern already in use in the docs. One of the advantages of the diff is point to the exact spot of the modification by keeping the context. Also, the amount of characters do be removed here is very minimal.

If there's a standard I should follow here please let me know and I'll change it. 😃

theZieger:
Totally fine. I guess one of the long term contributers could clarify this more easily.

flybayer:
Thanks for the effort @victorferreira, however I don't think we should add this to the docs. Firstly, for any production application it is **highly** recommended to use the same db in dev as in prod. Otherwise you will eventually run into weird bugs.

Also, my understanding from the Prisma team is that this feature is going to be removed.

As for the diff highlighting, you can use [line highlighting](https://github.com/blitz-js/blitzjs.com#code-syntax-highlights) instead of the diff to highlight specific lines.

victorferreira:
I'll close this PR, then.

Just out o curiosity, did the Prisma team document this anywhere?

flybayer:
I wasn't able to find anything with a quick search

PR #237, opened 2020-11-01
Not Anonymous arrow functions for Fast Refresh

matamatanot:
You will see the following First time users will be anxious.

```
warn  - ./pages/index.tsx
Anonymous arrow functions cause Fast Refresh to not preserve local component state.
```


flybayer:
Nice, thank you!!

PR #243, opened 2020-11-06
fix title font fallback

Talor-A:
fixes https://github.com/blitz-js/blitz/issues/1349. title falls back to serif, replace it with the values used by the body


flybayer:
Sweet, thanks!

PR #244, opened 2020-11-06
Add docs on deploying to DigitalOcean App Platform

### LINKED ISSUE #240: Add instructions for deploying on Digital Ocean App Platform
The title speaks for itself 😅

LoriKarikari:
I have created some documentation on how to deploy to DigitalOcean App Platform. I'm going to reread it and fix some stuff when needed so I'm leaving it as a draft for now.

Some notes:
- the build command I had to use is `npx blitz db migrate && npx blitz build` because otherwise, it throws an error that the command is not found. I tried adding a PATH variable and I was able to use the `blitz` command but the run command would fail so I left it out.
- the run command is `npx next start` because I was encountering the same issue as seen with Heroku where Blitz rebuilds. This is because some files are added during the build. I haven't figured out yet which files so this has to do for now. 

Closes #240


flybayer:
Yeah on CI you need to run `npm run blitz start`, `yarn blitz start`, or `npx blitz start` since blitz isn't installed globally. Or use package.json scripts

docs/deploy-digitalocean.mdx
```diff
@@ -0,0 +1,173 @@
...
+
+Create a new database.
+
+```bash
+doctl db create <name> --region <region> --engine pg --version 12
+```
+
+> **Note**: For this guide we are creating a PostgreSQL database but MySQL is also available on DigitalOcean.
+
+:::caution
```
Flavyoo, line 29: Nit: Is there an extra colon here? I checked the preview and it renders this as text.

docs/deploy-digitalocean.mdx
```diff
@@ -0,0 +1,173 @@
...
+doctl db list
+```
+
+Now you can get the connection string of to the connection pool.
+
+```bash
+doctl db pool list <database-id>
+```
+
+Add `&pgbouncer=true` to the end of the connection string and add it to `app.yaml`.
```
Flavyoo, line 126: Is it worth going into detail on why `&pgbouncer=true` is needed?

docs/deploy-digitalocean.mdx
```diff
@@ -0,0 +1,173 @@
...
+
+Now you can deploy the application.
+
+```bash
+doctl apps create --spec app.yaml
+```
+
+The first time you have to use `doctl` to connect your repository with App Platform. After that a new deployment will be triggered every time you push to your git repository.
+
+Now if you add the `.do` to your git repository values like `DATABASE_URL` and `SESSION_SECRET_KEY` will be exposed.
```
Flavyoo, line 162:
```suggestion
Now if you add the `.do` to your git repository values like `DATABASE_URL` and `SESSION_SECRET_KEY` they will be exposed.
```

Flavyoo (commented):
This is great documentation!

docs/deploy-digitalocean.mdx
LoriKarikari: Sure, I can add a note and a link to the Prisma docs.

docs/deploy-digitalocean.mdx
LoriKarikari: It looks like this locally preview

docs/deploy-digitalocean.mdx
LoriKarikari: Why is `they` needed here? English isn't my main language so I appreciate your feedback! I did forget `folder` after `.do` though. If you see other spelling or grammar mistakes feel free to correct them!

docs/deploy-digitalocean.mdx
Flavyoo: Nice! I guess it was because I was using the preview of the PR.

docs/deploy-digitalocean.mdx
Flavyoo: This is because it was not sure what will be exposed. But reading it again. I think I mis understood what this was saying.

Now i'm understanding this as: "If we add `.do` to my git repository, the `DATABASE_URL` and `SESSION_SECRET_KEY` will be exposed."

If this is true, we might need a comma after values instead of adding `they`. The comma will help separate the two parts of the sentence.

flybayer: This should be like Lori first wrote it.

flybayer:
Hey @LoriKarikari were you finished with this? (Still marked as Draft)

LoriKarikari:
> Hey @LoriKarikari were you finished with this? (Still marked as Draft)

Hi, I think some things changed since I created this because App Platform is fairly new. I'm not using it anymore since it didn't suit my needs so I forgot about it, sorry for that. I'll check if something has changed and remove the draft.

PR #245, opened 2020-11-07
Add model to type matrix

akirabaruah:
Documents the expected generated files for `blitz generate model` referenced elsewhere in these docs.


flybayer:
Ah, nice! Thanks!

PR #270, opened 2020-11-20
master is replaced by main in new github project

rap2hpoutre:
Since most people following the guide may have just created a fresh project hosted on Github, maybe we could suggest `main` instead of `master`.


flybayer:
Good Idea, thanks!

PR #273, opened 2020-11-25
Fix invalid syntax in API routes auth code example

williamkwao:
I was following the docs to authenticate through an API route and realized ```export default const`` in the code example is not valid after posting it in my code editor. This a small edit to clean it up. 

<table>
  <tr>
    <th>
     Before
     </th>
     <th>
     After
     </th>
  <tr>
  <tr>
    <td>
    Screen Shot 2020-11-25 at 4 13 58 PM
   </td>
<td>
Screen Shot 2020-11-25 at 4 17 32 PM 
  </tr>
</td>
</table>

### DIFF
--- a/docs/api-routes.mdx
+++ b/docs/api-routes.mdx
@@ -49,13 +49,13 @@ You can use the `getSessionContext` function to get the session of the user. Her
 ```ts
 import {getSessionContext} from "@blitzjs/server"
 
-export default const customRoute = async (req, res) => {
+export default async (req, res) => {
   const session = await getSessionContext(req, res)
   console.log("User ID:", session.userId)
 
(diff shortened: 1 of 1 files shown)


docs/api-routes.mdx
```diff
@@ -49,13 +49,13 @@ You can use the `getSessionContext` function to get the session of the user. Her
 ```ts
 import {getSessionContext} from "@blitzjs/server"
 
-export default const customRoute = async (req, res) => {
+export default async (req, res) => {
```
flybayer, line 52:
```suggestion
export default async function customRoute(req, res) {
```

flybayer:
Ah, good catch, thank you!

PR #274, opened 2020-11-30
Remove `manual` option occurrences

ugogo:
Remove `manual` option occurrences across queries pages since it seems to have been replaced by `enable`

image


flybayer:
Thank you! :)

PR #291, opened 2020-12-26
Fix for "Listing Choices" portion of Tutorial

phillipkregg:
The current code causes an error because the "FindManyQuestionArgs" type comes from Prisma.

This will help tutorial users by preventing this error and explicitly showing that they need to keep the Prisma imports from the code they are replacing.


flybayer:
Sweet, thank you!! @phillipkregg

PR #292, opened 2020-12-29
Move ZSH note to below first usage of blitz generate all

timReynolds:
Inline with the discussion on https://github.com/blitz-js/blitz/issues/1606#issuecomment-751389484 I've moved the note about ZSH to under the first place it would be needed to make it clearer for the reader.


flybayer:
Thank you @timReynolds !!

Issue #293, opened 2020-12-31
incorrect code in tutorial


linbudu599:
- tutorial:
```typescript
// app/questions/queries/getQuestions.ts
import { Ctx } from "blitz"
import db, { Prisma } from "db"
type GetQuestionsInput = Pick<Prisma.FindManyQuestionArgs, "where" | "orderBy" | "cursor" | "take" | "skip">
export default async function getQuestions(
  {where, orderBy, cursor, take, skip}: GetQuestionsInput,
  ctx: Ctx,
) {
  ctx.session.authorize()
  const questions = await db.question.findMany({
    where,
    orderBy,
    cursor,
    take,
    skip,
    include: {choices: true},
  })
  const count = await db.question.count()
  const hasMore = typeof take === "number" ? skip + take < count : false
  const nextPage = hasMore ? { take, skip: skip + take! } : null
  return questions
}

// app/questions/pages/questions/index.tsx
  const [{ questions, hasMore }] = usePaginatedQuery(getQuestions, {
    orderBy: { id: "asc" },
    skip: ITEMS_PER_PAGE * page,
    take: ITEMS_PER_PAGE,
  })
```

This will cause an error because props `questions`(`[{ questions, hasMore }]`) doesn't exist on type `(Question & { choices: Choice[]; })[]` (`return questions`).

should be like this:

```typescript
// app/questions/queries/getQuestions.ts
const count = await db.question.count()
const hasMore = typeof take === "number" ? skip ?? 0 + take < count : false
const nextPage = hasMore ? { take, skip: skip ?? 0 + take! } : null
return { questions, hasMore }
```

Nullish Coalescing in `skip ?? 0` is required to avoid typescript warnings.

flybayer:
Yes, you are right! Want to make a PR to fix it? :)

linbudu599:
> Yes, you are right! Want to make a PR to fix it? :)

Sure, but there's one more problem: the code in the tutorial seems to be different with code generated, e.g:

```typescript
// app/questions/pages/index.tsx

// docs
 <Link href="/questions/[questionId]" as={`/questions/${question.id}`}>
     <a>{question.text}</a> 
 </Link>

// generated
 <Link href={`/questions/${question.id}`}>
     <a>{question.text}</a>
 </Link>
```

should I directly change the tutorial content to keep up with code generated?

PR #295, opened 2021-01-01
fix: fix outdated & incorrect code example

linbudu599:
Closes: #293

- The `skip type definition` and `result return from getQuestions.ts` problem mentioned in the issue is my carelessness, content which generated from the latest version of BlitzJS works normally.

I found some other issues and  fix it, including: 
- `<Link />` component props:
- `include` options in getQuestion.ts
- keep up with latest generated contents

### DIFF
--- a/docs/tutorial.mdx
+++ b/docs/tutorial.mdx
@@ -522,6 +522,7 @@ export default async function getQuestions(
     orderBy,
     take,
     skip,
+    include:{ choices:true}
   })
 
   const count = await db.question.count()
@@ -537,18 +538,7 @@ export default async function getQuestions(
 }
 ```
 
-Now hop back to our main questions page in your editor, and we can list the choices of each question.
-
-Add `include` option in `app/questions/queries/getQuestion.ts` to fetch `Question` data with its choices info:
-
-```
-const question = await db.question.findFirst({
-  where,
-  include: { choices: true }, // highlight-line
-})
-```
-
-Then add this code beneath the `Link` in our `QuestionsList`:
+Now hop back to our main questions page in your editor, and we can list the choices of each question. And add this code beneath the `Link` in our `QuestionsList`:
 
 ```jsx
 <ul>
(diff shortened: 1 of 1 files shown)


docs/tutorial.mdx
```diff
@@ -513,32 +509,46 @@ export default async function getQuestion({where /* include */}: GetQuestionInpu
 import { Ctx } from "blitz"
 import db, { Prisma } from "db"
 
-type GetQuestionsInput = Pick<Prisma.FindManyQuestionArgs, "where" | "orderBy" | "cursor" | "take" | "skip">
+type GetQuestionsInput = Pick<Prisma.FindManyQuestionArgs, "where" | "orderBy" | "skip" | "take">
 
 export default async function getQuestions(
-  {where, orderBy, cursor, take, skip}: GetQuestionsInput,
-  ctx: Ctx,
+  { where, orderBy, skip = 0, take }: GetQuestionsInput,
```
linbudu599, line 515: It seems that I just copy the code when walking through the tutorials :)
so I just change the content to the latest generated code.

docs/tutorial.mdx
```diff
@@ -513,32 +509,46 @@ export default async function getQuestion({where /* include */}: GetQuestionInpu
...
 
-Now hop back to our main questions page in your editor, and we can list the choices of each question. Add this code beneath the `Link` in our `QuestionsList`:
+Now hop back to our main questions page in your editor, and we can list the choices of each question.
+
+Add `include` option in `app/questions/queries/getQuestion.ts` to fetch `Question` data with its choices info:
+
+```
+const question = await db.question.findFirst({
+  where,
+  include: { choices: true }, // highlight-line
```
linbudu599, line 547: We cannot use `question.choices` if we don't include choices relations in query operation.

docs/tutorial.mdx
```diff
@@ -640,7 +641,7 @@ export const Question = () => {
         where: {id},
         data: {votes: votes + 1},
       })
-      refetch()
```
flybayer, line 643: Sweet, thank you!!

Looks good except we need to keep this `refetch()` so the query get's updated after the mutation.

linbudu599: 🧐done

flybayer:
Perfect!

PR #297, opened 2021-01-05
tutorial db migration cosmetics

creimers:
I stumbled over this bit in the tutorial. My changes intend to make it clearer what to type and what output to expect.


flybayer:
Sweet, thank you!!

PR #309, opened 2021-01-09
Roll back the sentence to the original output

doi-t:
This PR reverts https://github.com/blitz-js/blitzjs.com/pull/297.

The sentence is actually a part of actual output of `blitz generate all question text:string choices:choice[]`. We need to run `blitz db migrate` after the second `blitz generate` command.


flybayer:
Ah, yeah 😅 thank you!

PR #311, opened 2021-01-09
V2 - extract out <Footer>

### LINKED ISSUE #261: V2 - Page Container
Create the Page Container component.

The PageContainer is a simple component that we wrap contents of pages with. It's main responsibility is to have a bounding width to prevent contents pages from extending beyond certain screen sizes or causing content in the pages to stretch too much.

Checkout the [Figma design](https://www.figma.com/file/KxLHA7cVPEy1fNE2bD8wWv/Blitz?node-id=1147%3A0). Take note of how there is a limit on the width of the page with margin to the left and right.

- [ ] Bounding width based on the figma design.

- [ ] Proper margin on left and right.

- [ ] Centered relative to the Layout component. #256

abuuzayr:
Implements a basic layout for the entire site, based on the current implementation. Serves as a starting point for further improvements. 

Should be merged first however so that any further work on the site (especially `index.js`) can use the layout. 

Closes #261


abuuzayr:
@Flavyoo mentioning you as requested.

flybayer:
@abuuzayr ah, I'm guessing you didn't see my comment in #261 from yesterday? 

I really appreciate the effort! But the homepage has a unique layout because of how the background colors change throughout the page and so its layout is unique and doesn't need a component (can stay inline).

**But**, what we **do** need is to extract the top nav and bottom nav into components because every page will use those.

abuuzayr:
@flybayer you're right! I did not see that.

Thanks for the explanation and direction. Will make changes and let you know when it's done!

abuuzayr:
@flybayer @Flavyoo I reverted the changes for `index.js`, no longer using a layout. Footer is now a component that should be used everywhere.

new-site/app/components/home/Footer.js
```diff
@@ -1,52 +1,81 @@
...
-      { title: "YouTube", href: "https://www.youtube.com/tailwindlabs" },
-    ],
-  },
-}
+import { Icon } from "@/components/home/Icon"
+import { LinkList } from "@/components/home/LinkList"
+import { NewsletterForm } from "@/components/home/NewsletterForm"
+import { IoLogoVercel } from "react-icons/io5"
 
 export function Footer() {
```
flybayer, line 7: Nice, looking good! One request, can you add an optional className prop on this component? Because it'll be needed sooner or later.

abuuzayr: Absolutely! Let me know if the changes I made were not what you were thinking of.

flybayer:
Nice, thanks!

PR #315, opened 2021-01-10
Polish the tutorial to make it more precise

doi-t:
While the tutorial shows the full example of `app/questions/queries/getQuestions.ts`, it shows a partial code of `app/questions/queries/getQuestion.ts`. So, I added imports and some declarations for the sake of consistency.

I also did some cleanups to make the tutorial more precise.


flybayer:
Awesome thank you!! :)

PR #322, opened 2021-01-13
V2 - Banner

### LINKED ISSUE #257: V2 - Banner
Create a banner component that can be used to make announcements on the site.

Requirements:

- [ ] Follow current design linked below

- [ ] Take the announcement content as a prop. Consider how we can store the announcement so it can be easily updated and translated. Maybe store it in an MDX file?

- [ ] Might be worth to render this banner component as part of the Page Layout component. This way we don't have to import it and render it in each page we make.

- [ ] Responsive, follows mobile design in Figma

Designs:
Figma: https://www.figma.com/file/KxLHA7cVPEy1fNE2bD8wWv/Blitz?node-id=1147%3A0

banner-component

abuuzayr:
Closes #257 

@Flavyoo


flybayer:
Nice thanks!

I just see one issue with the text not being centered on mobile:
image

new-site/app/components/Banner.js
```diff
@@ -0,0 +1,9 @@
+const Banner = ({ message }) => (
```
Flavyoo, line 1: should we add PropTypes for message?

new-site/app/components/Header.js
```diff
@@ -28,6 +29,7 @@ const Header = ({ className = "" }) => {
 
   return (
     <>
+      <Banner message="Blitz is now in beta!" />
```
Flavyoo, line 32: Nit: As an added bonus, is it possible this message can be stored in a file or in a way it can easily be updated and queried in case new messages have to be updated?

Flavyoo (approved):
@abuuzayr Thanks for adding this :)

Left a couple minor comments, but they are not blockers in my opinion.

new-site/app/components/Banner.js
abuuzayr: I thought of that, but I think instead of using PropTypes we should use Typescript on this repo. 

I set up an issue for your comments: https://github.com/blitz-js/blitzjs.com/issues/324

new-site/app/components/Header.js
abuuzayr: I agree with you, that messages like these should be quickly updated. My only thought is - while this isn't the only text on the site that can be managed elsewhere, its probably gets updated more than most. But, we're probably not going to offload this to a CMS to manage, that would be overkill, and I don't think thats what you meant. At some point this site will have to be translated, so this message later on would live in a `.json` file if we are using a library like [next-i18next](https://github.com/isaachinman/next-i18next).

Would that be better?

But while crafting this reply, I think I should also:

- [ ] Handle the scenario where the banner should not appear if there is no message passed
- [ ] Because the Banner is placed in the Header component, it could be that some pages might want to render the header without the Banner. So the Banner message should be passed from the Header props

flybayer: Hard coded here is fine for now :)

PR #331, opened 2021-01-18
Improve tutorial doc

mikeesto:
Hello :) I've been working through the tutorial. It's very useful. I just had some suggestions on some parts that I found a little confusing.

I also noticed on [line 659](https://github.com/blitz-js/blitzjs.com/blob/main/docs/tutorial.mdx#L659) there's a `refetch()` but it's not destructed from `useQuery`, and it's not included in the preceding code block (lines 603-625). Is that refetch needed? Currently the final step gives an error - refetch is not defined.

### DIFF
--- a/docs/tutorial.mdx
+++ b/docs/tutorial.mdx
@@ -27,7 +27,7 @@ From the command line, `cd` into the directory where you’d like to store your 
 blitz new mysite
 ```
 
-This should create a `mysite` directory in your current directory.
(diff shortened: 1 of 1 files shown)


flybayer:
Very nice, thank you!! And yes you are right about needing to add a step to get refetch:

```ts
const [question, {refetch}] = useQuery(getQuestion, {where: {id: questionId}})
```

mikeesto:
Thanks! I've added in the refetch step

flybayer:
Sweet, thank you!!

PR #342, opened 2021-01-27
Add docs for all the new `pipe` utilities

flybayer:
https://github.com/blitz-js/blitz/pull/1801


docs/resolver-server-utilities.mdx
```diff
@@ -0,0 +1,257 @@
...
+title: Server Utilities
+sidebar_label: Server Utilities
+---
+
+## `pipe.resolver`
+
+This is a functional utility that makes it easier and cleaner to write complex resolvers.
+
+- It reduces the need for explicit Typescript types
+- Validate input data with a zod schema with `pipe.zod()`
```
Flavyoo, line 11: Should we link to Zod in case some users don't know about it?

Flavyoo: I see it is linked below so maybe not needed here, but might be worth to add it here since it is the first mention.

docs/resolver-server-utilities.mdx
```diff
@@ -0,0 +1,257 @@
...
+
+- `...isAuthorizedArgs`: The same arguments as your [`isAuthorized` adapter](./authorization#isauthorized-adapters)
+
+### Returns
+
+A function to give `pipe.resolver` of type `(input, ctx: Ctx) => input`
+
+## `pipe.authorizeIf`
+
+This is like `pipe.authorizeIf` but the first argument is a callback function that determines whether the authorization will be enforced or not
```
Flavyoo, line 153: Nit: Period at the end.

Flavyoo (approved):

PR #343, opened 2021-01-27
Nextjs docs update

scisteffan:
Closes request by @flybayer


docs/incremental-static-regeneration.mdx
```diff
@@ -0,0 +1,711 @@
+---
+title: Incremental Static Regeneration
+sidebar_label: Incremental Static Regeneration
+---
+
```
flybayer, line 5: @scisteffan sweet, thank you! 

Looks like this file is duplicate. Content covered elsewhere. Can you confirm?

scisteffan: It's a new section from the nextjs docs that I thought would be helpful for Blitz users too who are wondering which methods to use for SSR, static generation or client side fetching. But yes, static regeneration is now at https://blitzjs-com-git-nextjs-docs-update.blitzjs.now.sh/docs/get-static-props#incremental-static-regeneration so it's ok to remove it.

flybayer: Ok let's remove it

PR #347, opened 2021-01-29
Mention API Routes in RPC spec

Skn0tt:


docs/rpc-specification.mdx
```diff
@@ -8,8 +8,11 @@ sidebar_label: RPC Specification
 This is documentation for using Blitz queries and mutations in a non-Blitz app. For example, from a mobile application or some other app.
 
 **For using queries and mutations in your app, refer to [Using Queries](./query-usage.mdx) and [Using Mutations](./mutation-usage.mdx).**
+
+**If you want to access your app e.g. from an external API, take a look at [API Routes](./api-routes.mdx) instead.**
```
Zeko369, line 12:
```suggestion
**If you want to access your app e.g. as an external API for a mobile app, take a look at [API Routes](./api-routes.mdx) instead.**
```

Zeko369 (approved):

docs/rpc-specification.mdx
Skn0tt: This PR was made in response to https://discord.com/channels/802917734999523368/804073762496184341/804708509459415070, and they didn't mention anything about a mobile app. We should put this more broadly, I think.

Zeko369: Regardless I'd make it as opposed to from because with API routes you're using your blitz app 'as' an external api.

From sounds like you're requesting data from the an external api not your app being that external api

PR #357, opened 2021-02-11
Update authorization.mdx

osirvent:
`ctx.session.$isAuthorized("customer")` should be `true` with User logged in with 'customer' role


flybayer:
Yes, thank you!

PR #361, opened 2021-02-13
Mention supported Node.js version, and explicitly mention Node.js 15…

arenddeboer:
… being unsuported.

See https://github.com/blitz-js/blitz/issues/1668 for a discussion on why Node.js 15 is currently unsupported.


flybayer:
Hey thanks @arenddeboer! However Node 15 is supported with latest canary release as of https://github.com/blitz-js/blitz/pull/1859 :)

PR #363, opened 2021-02-15
Fix dateTime example on Scalar Fields

fmilani:
The example for creating a model with a `dateTime` field doesn't have the capital T. If the model is created with the lowercase t, the prisma migration fails.


flybayer:
Yes, thank you very much!!

PR #364, opened 2021-02-15
Docs bugs

abuuzayr:
Fix bugs for docs pages:

- Dont allow scroll when menu is open
- Copyright font color
- Short circuit logic for topics where there are no topics to drop down
- Back to Docs center align


flybayer (approved):
Thank you!

PR #379, opened 2021-02-21
Fix input param for CreateQuestion

andreadellacorte:
I noticed that the parameter for CreateQuestion doesn't match the example - Proposing a fix


flybayer:
Nice, thank you!!

PR #380, opened 2021-02-22
Fixed handles

### LINKED ISSUE #377: Fixed handles (necessary for i18n)
To prevent broken handles when doing i18n, we should implement something similar to what React does in their documentation

English:
```md
## UX for Unauthorized Users {#ux-for-unauthorized-users}
```

Spanish:
```md
## UX para Usuarios no autorizados {#ux-for-unauthorized-users}
```

JuanM04:
Fixes #377

- Added slugs to the plugin (`remark/withTableOfContents.js`)
- Added documentation
- Created script to automate the slugify
- Slugified all the documents


README.md
```diff
@@ -38,3 +38,39 @@ const highlighted = true
...
+```md
+### Install Blitz {#install-blitz}
+```
+
+There are a few more rules:
+
+- All titles smaller than H3 (or the ones with more than `###` at the start) must not have a handle.
+- The handles are entirely in lowercase and only the 26 letters of the English alphabet and numbers are allowed.
+- Instead of a space, use `-`.
+- There must not be two identical handles in the same document. If you have two identical titles, add a number at the end in order of appearance (not importance)
```
flybayer, line 62: Sweet thank you @JuanM04!!

Question, does the build throw an error if <H3 has a handle? And does build throw an error if there is a duplicate?

I'm hoping the answer is yes :)

JuanM04: It throws an error if there are two identical in the same document.
It does _not_ throw an error if a <H3 has it... but it doesn't parse it either. Will result something like this:

```
### Pages {#pages}
#### Context {#context}
```

Would turn out to:

```
Pages
Context {#context}
```

Maybe I could add a check. Do you want me to check for a <H3?

flybayer: Ok yeah if that will turn into a bug, then let's throw an error to prevent it

flybayer: Also you have a merge conflict from your other PR I merged

JuanM04: 👍

PR #387, opened 2021-02-24
Languages and Translations entry

JuanM04:
I added a very simple languages page. It's only meant to track the progress of each translation (for now)

Also, I added a entry in the docs with a bit of information for everyone who wants to translate Blitz. It's meant to be used in this situations:

> — Hey, how can I help translating Blitz?
> — https://blitzjs.com/docs/translations

### DIFF
--- a/app/navs/documentation.js
+++ b/app/navs/documentation.js
@@ -62,6 +62,7 @@ export const documentationNav = [
       pages["contributing"],
       pages["maintainers"],
       pages["code-of-conduct"],
+      pages["translations"],
     ],
   },
   {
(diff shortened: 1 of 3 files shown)


flybayer:
Sweet, thanks!!!

Maybe add "The React documentation is available in the following languages:" under the "Language" header like react has, which helps explain the page I think.

Also you have an min-height issue:

image

flybayer:
You're a genius by the way. I love the automation here! :D

JuanM04:
> Also you have an min-height issue

Yes, and the footer is behaving weirdly. I'm working on it

JuanM04:
@flybayer Does the min-height error persists?

PR #388, opened 2021-02-25
style: update search button styles

Kosai106:
When hovering the search button using the bright theme, the text colour would remain black, making it hard to read.

example

### DIFF
--- a/app/components/Search.js
+++ b/app/components/Search.js
@@ -67,7 +67,7 @@ export function Search({ className = "" }) {
         type="button"
         ref={searchButtonRef}
         onClick={onOpen}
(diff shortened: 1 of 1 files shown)


JuanM04:
That's weird. I see that working fine. Could you share your browser and its version? Also, it would be great if you could share a short GIF showing the bug, but it's not necessary

Kosai106:
@JuanM04 I've added a gif that highlights the issue.

I should have clarified that it works fine on the homepage. The issue occurs on the docs.

JuanM04:
Ahhh, I see it now.

Thank you!

JuanM04 (approved):

Issue #24, opened 2020-04-19
Initial Blitz cli commands


robdrosenberg:
Create alpha documentation for getting blitz up and running locally and cli commands.

- [ ] Highlight blitz cli commands
- [ ] Initial Blitz project setup

flybayer:
I have an alpha guide here with everything needed for the alpha (I think): https://github.com/blitz-js/blitz/pull/131

So that's a good starting point for this issue.

coreybrown89:
@flybayer thanks - this is super helpful.

PR #394, opened 2021-03-01
Add type annotations to api-routes

aditsachde:
Should lines #24-25 be changed? I don't think there's a better place to link though.


flybayer:
Hey thanks! And yes you can change those lines to be like this:

```
- `req: BlitzApiRequest`: ...
```

aditsachde:
Updated!

JuanM04:
One last change! Change `js` to `ts` in the codeblocks

PR #395, opened 2021-03-02
feat(docs): add impersonation guide

Kosai106:
I took the liberty of also updating the guide to make use of the new resolvers.

My writing skills are not the best so I didn't add too much additional content from the original guide so feel free to come with suggestions for improvements if necessary.


flybayer:
Sweet thanks!! You need to rename `impersonation.tsx` to `impersonation.mdx` :)

And then I'll review further

Kosai106:
@flybayer you're right 🤦‍♂️ Old habits die hard.

JuanM04 (approved):
Very nice! I wouldn't used assert, but it works

Kosai106:
> Very nice! I wouldn't used assert, but it works

It's a left over from the original guide although maybe it makes sense to replace `assert` and `logger` as I don't think they come with Blitz projects by default. Let me know what you think and I'll update the PR accordingly.

JuanM04:
Yes, the paths are wrong... maybe replace them by a simple "if not exist, then throw" and a "console.log". For simplicity 😄

JuanM04 (approved):

app/pages/docs/impersonation.mdx
```diff
@@ -0,0 +1,157 @@
+---
+title: How To Impersonate Other Users
+sidebar_label: How To Impersonate Other Users
+---
+
+Impersonation, also called personification is often a critical tool for SaaS applications as it allows you to view the application from the perspective of a specific user without knowing their credentials.
```
flybayer, line 6:
```suggestion
Impersonation is often a critical tool for SaaS applications as it allows admins to view the application from the perspective of a specific user without knowing their credentials.
```

"personification" is something totally different :)

flybayer:
Awesome, thanks so much!! Now we can remove it from the wiki

PR #396, opened 2021-03-02
Print width of 74 in all documents

### LINKED ISSUE #389: Remove horizontal scroll in code examples by reformatting code blocks with smaller line width
First off: the docs are beautiful and I as an absolute blitzjs beginner have found everything i needed within seconds. Good job on that!

Little nitpick though:
image

In most code examples i got a horizontal scrollbar making it hard to read them.
I would imagine that if the right column would use the full space (till underneath the social icons) this would be less problematic.

Wdyt? Any other suggestions? 
I'd be happy to work on that.

JuanM04:
Fixes #389

Also, changes the prettier config to match the one used in the `blitz` repo


flybayer:
Poor translators 😅

PR #402, opened 2021-03-09
Add blitz export page

Skn0tt:


JuanM04:
@Skn0tt What do you think about either adding a link to Next.js documentation or copy it to this page?
https://nextjs.org/docs/advanced-features/static-html-export

Skn0tt:
Good idea! Added a link.

JuanM04 (dismissed):

flybayer:
@Skn0tt the goal is to not have any links to next.js documentation, because some things can be different.

If we don't have the that static export page, then we should copy that content and add it to our docs

PR #404, opened 2021-03-10
Fix render.yaml example of deploy-render `Without database`

the-red:
There was unwanted properties that caused the build to fail,
so I removed them.


flybayer:
Yes, thank you!! Sorry about that. Not sure how that example got messed up.

the-red:
@flybayer 

Thanks for the quick merge.
I forgot to copy the error message, but I got two different errors.

* The database "blitzapp-db" does not exist.
* The property name "startCommand" is duplicated.

PR #411, opened 2021-03-18
Apply code replacement for // highlight-line

### LINKED ISSUE #410: // highlight-line syntax no longer highlights a line in V2
When we moved to V2, the previous `// highlight-line` syntax to highlight line in code blocks no longer works.

abuuzayr:
Fixes #410 

This is an ugly hack, I must apologise. What we should do next is to find a different way to write the `// highlight-line`, `// highlight-start` and `// highlight-end` for syntax highlighting if we want to go about doing this better.

### DIFF
--- a/remark/utils.js
+++ b/remark/utils.js
@@ -118,7 +118,6 @@ module.exports.highlightCode = function highlightCode(code, prismLanguage) {
   code = code
     .replace(/[^\S\r\n]*(?=\/\/ highlight-start)/g, "")
     .replace(/[^\S\r\n]*(?=\/\/ highlight-end)/g, "")
-    .replace(/.*\/\/ highlight-line$/gm, (text) => `highlightlinestart ${text} highlightlineend`)
   let highlighted = Prism.highlight(code, grammar, prismLanguage)
 
   return highlighted
@@ -129,10 +128,8 @@ module.exports.highlightCode = function highlightCode(code, prismLanguage) {
     .replace(/\w*\/\/ ?highlight-start\w*/g, "")
     .replace(/\w*\/\/ ?highlight-end\w*/g, "")
     .replace(
-      /(highlightlinestart )[\S\s]*?( highlightlineend)/g,
-      (text) => `<div class="token highlighted highlighted-line">${text}</div>`,
+      /^.*\/\/ ?highlight-line.*$/gm,
+      (text) => `<div class="token highlighted">${text}</div>`,
     )
-    .replace(/highlightlinestart/g, "")
-    .replace(/highlightlineend/g, "")
-    .replace(/\/\/ highlight-line/g, "")
+    .replace(/<span class="token comment">\/\/ ?highlight-line<\/span>/g, "")
 }
--- a/remark/utils.js
+++ b/remark/utils.js
@@ -115,10 +115,10 @@ module.exports.highlightCode = function highlightCode(code, prismLanguage) {
     console.warn(`Unrecognised language: ${prismLanguage}`)
     return Prism.util.encode(code)
   }
-  code = code.replace(
-    /.*\/\/ highlight-line$/gm,
-    (text) => `highlightlinestart ${text} highlightlineend`,
-  )
+  code = code
+    .replace(/[^\S\r\n]*(?=\/\/ highlight-start)/g, "")
+    .replace(/[^\S\r\n]*(?=\/\/ highlight-end)/g, "")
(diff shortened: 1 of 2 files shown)


abuuzayr:
current docs: 

Screen Shot 2021-03-18 at 4 41 39 PM

updated:

Screen Shot 2021-03-18 at 4 41 19 PM

Do let me know the right color to use also!

remark/utils.js
```diff
@@ -115,13 +115,24 @@ module.exports.highlightCode = function highlightCode(code, prismLanguage) {
...
     .replace(/\w*\/\/ ?highlight-start\w*/g, "")
-    .replace(/\w*\/\/ ?highlight-end\w*/g, "")
+    .replace(/\w*\/\/ ?highlight-end\w*/gm, "")
+    .replace(
+      /(highlightlinestart )[\S\s]*?( highlightlineend)/g,
+      (text) => `<div class="token inserted">${text}</div>`,
+    )
+    .replace(/highlightlinestart/g, "")
+    .replace(/highlightlineend/g, "")
+    .replace(/\/\/ highlight-line/g, "")
```
JuanM04, line 137: Thank you very much! If you could, use `highlighted` instead of `inserted`.

And btw, what do you think about doing something like this?

```js
.replace(
  /^.*\/\/ ?highlight-line.*$/gm,
  (text) => `<div class="token highlighted">${text}</div>`,
)
.replace(/<span class="token comment">\/\/ ?highlight-line<\/span>/g, "")
```

abuuzayr: Thanks for the review @JuanM04!

I have updated it to use `highlighted` instead of `inserted`.

I also had to tweak some styles that gave highlighted blocks unneeded spacing. 

As for your suggestion, that would be cleaner but i think it would not be able to work. If we apply the replacement to HTML before Prism does the highlighting, Prism will highlight the markup we replaced as code. If we apply the replacement after Prism does the highlighting, the code will look like:

```
...
<span class="token">some code</span>
<span class="token comment">// highlight-line</span>
...
```

and therefore only the second line will be wrapped in `<div class="token highlighted">`

JuanM04: > I also had to tweak some styles that gave highlighted blocks unneeded spacing.

Nice!

> As for your suggestion, that would be cleaner but i think it would not be able to work.

I thought Prism works with `<pre/>` instead of `<br/>`...

abuuzayr: > I thought Prism works with `<pre/>` instead of `<br/>`...

I'm not familiar around Prism, so that's probably true in a usual setting. However, from what I understand on how it works for our docs: 

1. mdx files are processed 
2. remark extracts the text from code blocks (at this stage, there's no markup - just pure text delimited by `\n`)
3. `Prism.highlight()` is applied on the text and spits out HTML, which are separated by *tokens* wrapped in a `<span/>` tag

Yes, `Prism.highlight()` would produce a `<pre/>` block, but once the text is *tokenized* we don't have a clear idea from the HTML on where a line break occurs (there are no `<br/>`s) which is why I used an arbitrary `highlightlinestart/end` to mark it and remove it later.

So sorry if this does not make sense, I'm not too sure on how to explain it better.

JuanM04: Makes sense, but because is delimited by `\n`, we can use the handy `/m` that only checks each line. You can see the output if you `console.log(highlighted)`

abuuzayr: You're absolutely right! Thanks for making the code cleaner. No idea why the first time I tried it only the `// highlight-line`'s span was highlighted. 

Anyway, I have updated the code as per your suggestion. 

Thank you!

JuanM04: Thank _you_ :blush:

remark/utils.js
```diff
@@ -115,13 +115,21 @@ module.exports.highlightCode = function highlightCode(code, prismLanguage) {
     console.warn(`Unrecognised language: ${prismLanguage}`)
     return Prism.util.encode(code)
   }
+  code = code
+    .replace(/[^\S\r\n]*(?=\/\/ highlight-start)/g, "")
+    .replace(/[^\S\r\n]*(?=\/\/ highlight-end)/g, "")
```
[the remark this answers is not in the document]
abuuzayr: this removes whitespace ahead of `// highlight-[start,end]`, otherwise I can’t seem to apply a consistent styling for cases where there is whitespace and cases where there isn’t.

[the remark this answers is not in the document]
JuanM04: Awesome! Thank you!

JuanM04 (approved):

PR #415, opened 2021-03-18
chore: Fix "GitHub" capitalisation

franky47:
Github -> GitHub


JuanM04:
Thank you for noticing that. I'm sure there are other wrong capitalisations in the docs. Would you like to fix them too? (you can find them with `grep -i "github"` or with VSCode search)

JuanM04:
Thank you for noticing that!

JuanM04 (approved):

PR #416, opened 2021-03-19
Optional chore: Add and use @tailwindcss/jit

abuuzayr:
Tailwind just released a spanking new experimental library for just-in-time compilations a few days ago: https://blog.tailwindcss.com/just-in-time-the-next-generation-of-tailwind-css

For me who's using an underpowered laptop, developing on blitzjs.com was quite a pain as any change I made would trigger a CSS recompilation and my browser would struggle to apply the large css bundle as we enabled quite a few variants (perhaps we don't even use all the enabled variants, but since V2 was a fork off the tailwind website.. we didn't look into removing the unused variants)

While the library is experimental, I think it is sufficiently ready and will bring a lot of benefits by this simple small change. But of course, this is optional and we don't have to apply it if we don't want to :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
     "@mdx-js/loader": "^1.6.22",
     "@next/bundle-analyzer": "^10.0.8",
     "@svgr/webpack": "^5.4.0",
+    "@tailwindcss/jit": "0.1.3",
     "@tailwindcss/typography": "0.4.0",
     "alex": "9.1.0",
     "autoprefixer": "^10.2.5",
--- a/postcss.config.js
+++ b/postcss.config.js
@@ -1,3 +1,3 @@
 module.exports = {
-  plugins: ["tailwindcss", "postcss-nested", "autoprefixer"],
+  plugins: ["@tailwindcss/jit", "postcss-nested", "autoprefixer"],
 }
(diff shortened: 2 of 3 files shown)


JuanM04:
I did a quick benchmark and the results were amazing!
When building the whole site, before it took 1m15s and now it takes 55s (-25%).
But, when running `yarn dev`, the time it takes to the first render went from 26s to 7s (-75%!!!!!!)

abuuzayr:
hahah thanks for the benchmark results @JuanM04 ! i wanted to do it too but i had no idea how.. now thats quantifiable

JuanM04:
> i wanted to do it too but i had no idea how

It's not the best method, but when you run `yarn build` it says "Done in XX at the end", and to test the time to first render I just used a stopwatch :rofl:

JuanM04 (approved):

abuuzayr:
> > i wanted to do it too but i had no idea how
> 
> It's not the best method, but when you run `yarn build` it says "Done in XX at the end", and to test the time to first render I just used a stopwatch 🤣

if it works, it works! and i should stop overthinking about “the best way” 😝

flybayer:
woohoo!!

PR #417, opened 2021-03-21
Fix imports in api route example

FarazPatankar:
Very small but useful change here. The request parameters are typed but the types aren't imported.


JuanM04 (approved):
Thank you!

PR #418, opened 2021-03-21
Add docs for Error Handling components

ericvicenti:
Just putting this here for posterity because [I was confused](https://github.com/blitz-js/blitz/issues/2131) by this as a newcomer to blitz


flybayer:
Thank you @ericvicenti!!!

PR #421, opened 2021-03-25
render.yaml fix to properly deploy to Render

gstranger:
When following the above example to deploy to render I received an error that I needed to pass the SESSION_SECRET_KEY. In this change I leverage Render's IaC capability to generate this secret at runtime so new users can quickly deploy without errors.


gstranger:
Good call @kevinlangleyjr , I've opened a [PR](https://github.com/blitz-js/blitz/pull/2148) as well in the main blitzjs repo.

flybayer:
thank you!

PR #425, opened 2021-03-30
Remove redundant steps from `deploy-render.mdx`

konradkalemba:
I just deployed my blitz app to render.com and while I had been following official instructions I had noticed that two steps can be removed.

#421 introduced automatic `SESSION_SECRET_KEY` environment variable creation which makes steps about adding it manually not necessary anymore.


flybayer:
Thanks @konradkalemba!!

PR #429, opened 2021-04-06
Remove mentions of `as` (no longer required)

Skn0tt:


Skn0tt:
> `as` - Optional decorator for the path that will be shown in the browser URL bar. Before Next.js 9.5.3 this was used for dynamic routes, check our previous docs to see how it worked
> ~ https://nextjs.org/docs/api-reference/next/link

app/pages/docs/link.mdx
```diff
@@ -58,25 +58,20 @@ such cases instead.
...
+<Link href="/post/abc">
   <a>First Post</a>
 </Link>
 ```
 
-`href` is a file system path used by the page and it shouldn't change at
-runtime. `as` on the other hand, will be dynamic most of the time
-according to your needs. Here's an example of how to create a list of
-links:
-
```
JuanM04, line 74: Maybe you could leave "Here's an example of how to create a list of links" or something similar

JuanM04 (changes requested):
Awesome!

JuanM04 (approved):

PR #435, opened 2021-04-10
Error in tutorial

simonelnahas:
I got an error from running:
```sh
blitz generate resource choice text votes:int:default=0 belongsTo:question
```
the fix:
 `blitz prisma format


flybayer:
Thank you!

PR #439, opened 2021-04-12
Fix the first two issues of #436

laubonghaudoi:
The instructions of data migration is not clear in the tutorial, as described in https://github.com/blitz-js/blitzjs.com/issues/436

The last two issues are remained to be fixed.

### DIFF
--- a/app/pages/docs/tutorial.mdx
+++ b/app/pages/docs/tutorial.mdx
@@ -276,8 +276,9 @@ blitz generate resource choice text votes:int:default=0 belongsTo:question
 ```
 If you get an error run `blitz prisma format`
 
-Again, hit **Enter** when prompted to run the migration and enter a name
-for the migration.
+Note that this doesn't require a database migration because we haven't added the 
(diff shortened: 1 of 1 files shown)


app/pages/docs/tutorial.mdx
```diff
@@ -316,9 +317,36 @@ model Question {
...
+Your database is now in sync with your schema.
+```
+
+And then run:
+
+```sh
+blitz prisma generate
+```
+
+to update prisma client for that schema change.
```
flybayer, line 349: This step is not needed because `blitz prisma migrate dev` already generates the client too

flybayer (commented):
Thank you!

app/pages/docs/tutorial.mdx
laubonghaudoi: Thanks for pointing out! I have removed that part. Please check if it is ready to merge.

flybayer:
yay!

PR #451, opened 2021-04-19
Update docs examples to remove anonymous fn exports 

Roesh:
There are locations in the docs where we show example functions that are unnamed default exports.

When I created a custom api based on this example, my ES Lint complained:

`error    Assign arrow function to a variable before exporting as module default  import/no-anonymous-default-export`

The fix is to do as the linted advises. I have updated the locations in the docs that I found have anonymous default exports in this PR


JuanM04:
Thanks! Now that you bring this up, maybe you should name all of them as `handler` and write not like this:

```ts
const hello = (req: BlitzApiRequest, res: BlitzApiResponse) => {
  res.statusCode = 200
  res.setHeader("Content-Type", "application/json")
  res.end(JSON.stringify({name: "John Doe"}))
}
export default hello
```

but like this:

```ts
const handler: BlitzApiHandler = (req, res) => {
  res.statusCode = 200
  res.setHeader("Content-Type", "application/json")
  res.end(JSON.stringify({name: "John Doe"}))
}
export default handler
```

JuanM04:
Thank you! What do you think about naming them all as "handler"? (That's how Next.js calls them, I don't know why Blitz doesn't export a `BlitzApiHandler`)

JuanM04:
I just opened a PR about using `BlitzApiHandler`: https://github.com/blitz-js/blitz/pull/2239

Roesh:
Cool! Yeah, I can update them all to handler instead

JuanM04 (approved):

PR #452, opened 2021-04-19
Security Policy

JuanM04:
Draft for the security policy.


flybayer:
Excellent, thank you

JuanM04:
The only thing to do is creating a PGP pair

flybayer:
@JuanM04 what does that take? Never done that before

JuanM04:
@flybayer it's basically a way to encrypt messages, like when you encrypt your commits and they appear in GitHub as "Verified". You can read more about it here: https://www.digitalocean.com/community/tutorials/how-to-use-gpg-to-encrypt-and-sign-messages

The steps are for Ubuntu, but I think you can install `gpg` with Homebrew in Mac; I don't know how to do it in Windows

PR #454, opened 2021-04-23
add info for proxy-support https://github.com/blitz-js/blitz/pull/2264

mubaidr:
This adds documentation for proxy-support added by the following pull request. 
https://github.com/blitz-js/blitz/pull/2264

### DIFF
--- a/app/pages/docs/blitz-config.mdx
+++ b/app/pages/docs/blitz-config.mdx
@@ -144,8 +144,8 @@ module.exports = {
 
 ### Proxy support {#proxy-support}
 
-Proxy support is enabled automatically for recipe install if `http_proxy`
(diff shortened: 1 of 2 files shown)


app/pages/docs/blitz-config.mdx
```diff
@@ -140,6 +142,24 @@ module.exports = {
 }
 ```
 
+### Proxy support {#proxy-support}
+
+Proxy support is enabled automatically for recipe install if `http_proxy`
+environment variable is present. You can also set proxy option using:
```
flybayer, line 148: Can you also add the other two environment variables that we use?

mubaidr: Sure. Added `https_proxy` in the description too.

PR #455, opened 2021-04-24
Recommend customTsParser for building a new recipe

tmcw:
This recommendation for writing a new recipe could be improved: the docs say that you can/should use a direct import of jscodeshift. You can do this, but it'll fail pretty soon: what a recipe really provides is a program parsed with the customTsParser, not a program parsed with the default parser.


flybayer:
Thanks!!

PR #456, opened 2021-04-25
Admonitions --> Cards

JuanM04:
Replaced admonitions with a custom component. This lets us customize it as we like and we no longer need to use `<!-- prettier-ignore -->`


flybayer:
🚀 🚀

PR #460, opened 2021-04-27
Extend "run postgres via docker" section

antonykamp:
Hey,

more small changes: 
I've added some information about the location of `DATABASE_URL` and that the user has to run `docker-compose` before `blitz prisma migrate `.


flybayer:
Thank you!!

PR #461, opened 2021-04-28
Add `scroll-behavior: smooth;` to main.css

dawnofmidnight:
This pr is rather simple. It simply makes the scrolling when you scroll to another part of a page using an id button much smoother. For example, when you use the "Back to top" button in the footer, it won't jump up so quickly, it'll instead scroll. This is useful because it will surprise people less and it'll seem cleaner and less choppy too.

### DIFF
--- a/app/core/styles/main.css
+++ b/app/core/styles/main.css
@@ -11,6 +11,7 @@
 
 html {
   font-family: "Libre Franklin", sans-serif;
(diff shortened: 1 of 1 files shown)


flybayer:
Sweet, thank you!!

dawnofmidnight:
Glad I could help!

PR #469, opened 2021-05-10
Toggling text and removing outline

### LINKED ISSUE #463: Dark mode text not changing on toggling theme
When we click on toggle button for changing the theme on the website, the theme changes but the text in both cases always reads **Dark**,  signifying dark mode is applied, which might be confusing. 

Text-in-dark-mode

**Text in Dark Mode**

Screenshot 2021-05-05 at 10 29 28 PM

**Text in Light Mode**

I believe we can change the text on click to either **Light** or **Dark** depending on the theme applied. I think we can also remove the border around the toggle and the text when clicked, seems a bit weird to me, if it is ok with you. Let me know.

Hardik3296:
changes so that text toggles depending on theme selected. Also removed outline around toggle that stays when clicked.
resolves #463

### DIFF
--- a/app/core/components/DarkModeToggle.js
+++ b/app/core/components/DarkModeToggle.js
@@ -18,7 +18,7 @@ const DarkModeToggle = ({className}) => {
   return (
     <button
       onClick={switchTheme}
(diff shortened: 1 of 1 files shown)


flybayer:
Thanks! :)

Issue #470, opened 2021-05-11
Improve site performance


Roesh:
Brandon used lightouse and found that we could improve performance:

https://lighthouse-dot-webdotdevsite.appspot.com//lh/html?url=https%3A%2F%2Fblitzjs.com

[app/pages/index](https://github.com/blitz-js/blitzjs.com/blob/e6f46efd1502e8c98c137c69cf7fac4653466a5d/app/pages/index.js#L227) loads several github avatars using a plain img tag. In order to use an Image component, we need to add the domains we are receiving the images from to module.exports

```
images: {
    domains: [
      "avatars.githubusercontent.com",
      "avatars0.githubusercontent.com",
      "avatars1.githubusercontent.com",
...
```
Doing domains: [ *.githubusercontent.com"], but that doesn't work. I tried to see if github has like a certain number of these like maybe 5 max, but they don't say the number:

> To host your images, GitHub uses the open-source project Camo. Camo generates an anonymous URL proxy for each image which hides your browser details and related information from other users. The URL starts https://<subdomain>.githubusercontent.com/, with different subdomains depending on how you uploaded the image. 

[Info about avatar storage](https://docs.github.com/en/github/authenticating-to-github/about-anonymized-image-urls#troubleshooting-issues-with-camo)

There are a few next steps I see here:
1) Add regex support to the images.domains array so we can load any subdomain easily
2) Reach out to the github team to ask about how many subdomains there could be max, and if their names may change.
3) @MrLeebo suggested replacing `<img>` with `<Image>` will **not** really benefit things for githubusercontent.com. github is almost certainly already optimizing the images and hosting them on CDNs, so our images are probably better off as `<img loading="lazy">` elements
4) @MrLeebo also mentioned that if we could make improvements to github's own hosting, we could do something like 
```<Image src={`/avatar?src=${encodeURIComponent(contributor.avatar_url)}`} />```
and set up a custom route to proxy the githubusercontent domain so you don't have to whitelist each of them

I am going to start with the third option of lazy loading images and seeing if that improves our performance sufficiently, and work on the fourth option if it does not.

5) I reviewed the react-player npm package and have added changes to improve performance. The only drawback is that it takes two clicks to play the video. Initially, there is an image with the video thumbnail. A user's first click loads the actual youtube player, and the second click actually starts the video.

Roesh:
[...]

PR #474, opened 2021-05-11
Various changes to improve performance

Roesh:
Try to improve page performance

Tracked in #470


app/core/components/home/VideoPlayer.js
```diff
@@ -1,4 +1,4 @@
-import ReactPlayer from "react-player"
+import ReactPlayer from "react-player/lazy"
```
Roesh, line 1: From the (react-player)[https://www.npmjs.com/package/react-player] documentation:
> If your build system supports import() statements, use react-player/lazy to lazy load the appropriate player for the url you pass in. This adds several reactPlayer chunks to your output, but reduces your main bundle size.

app/core/components/home/VideoPlayer.js
```diff
@@ -9,6 +9,7 @@ const VideoPlayer = ({url, className = ""}) => {
         width="100%"
         height="100%"
         controls={true}
+        light={true}
```
Roesh, line 12: From the (react-player)[https://www.npmjs.com/package/react-player] documentation:
> The light prop will render a video thumbnail with simple play icon, and only load the full player once a user has interacted with the image.

app/core/components/home/VideoPlayer.js
Roesh: It also says
> If you only ever use one type, use imports such as react-player/youtube to reduce your bundle size. See config keys for all player keys.
so if we only plan on embedding youtube vids, we could use that instead

JuanM04:
Perfect!
Have you tried using `<Image />`? Is there a performance benefit by using it?

Roesh:
> 
> 
> Perfect!
> Have you tried using `<Image />`? Is there a performance benefit by using it?

Yes, I believe the should be, the typical github image is 460 pixels x 460 pixels and were serving 110 pixels, so we should get a benefit. The issue is that I'm having an error now even though I add the github domains to blitz.config.js; it doesn't seem to be picking that up. I can push those changes and check if the vercel preview is OK

JuanM04:
Well, it doesn't seem to throw an error. The only thing is that the CSS is messed up, but that should be a simple fix

Roesh:
Yep! I didn't expect that lol. Let me try to get my dev environment working and I can make sure the sizing is right

JuanM04:
Perfect, thank you!

Roesh:
So, that works, but it looks like our sizing configuration could be improved. We're still serving the original size, and I notice that there are sizes that are larger present, but not smaller sizes. The smallest size in srcset is 640w, so I'm going to see if this can be made smaller to 128w

JuanM04:
Ok. I don't remember right now (and I'm going to be busy in a few minutes), but maybe we are _only_ using 110px. If that's the case, we could use a fixed size. Could you check that?

Also, btw, could you change the userTrack sponsor image URL to this: `https://raw.githubusercontent.com/blitz-js/blitz/canary/assets/usertrack.png`?

Roesh:
> 
> 
> Ok. I don't remember right now (and I'm going to be busy in a few minutes), but maybe we are _only_ using 110px. If that's the case, we could use a fixed size. Could you check that?
> 
> Also, btw, could you change the userTrack sponsor image URL to this: `https://raw.githubusercontent.com/blitz-js/blitz/canary/assets/usertrack.png`?

Hmm, I don't believe so, on mobile, this changes to 62 px

Roesh:
The only thing that I think may be added here is an error boundary around the image tag just in case we try to get an image from a domain that isn't listed. Actually I'll work on that, might take until end of day since I'll be working on some other stuff, so you can review and merge after that is finished, Juan

JuanM04 (approved):
:rocket:

JuanM04:
> The only thing that I think may be added here is an error boundary around the image tag just in case we try to get an image from a domain that isn't listed

There is only one image that isn't listed, but it also throws and 401 (it doesn't appear in the README), so... we'll leave it for now

PR #478, opened 2021-05-18
Add doc page for eslint config and fix typo

Roesh:
I added a page for eslint configuration, borrowing from the [create-react-app docs](https://create-react-app.dev/docs/setting-up-your-editor/#experimental-extending-the-eslint-config) (Which have been linked for further reference, since we use it)

I am unable to test the documentation of the app because due to the following error which is thrown when I click on the documentation this url -> http://localhost:3000/docs/get-started

The prop `href` expects a `string` or `object` in `<Link>`, but got `undefined` instead.

There are several link components in the app directory, and many of them use parameters for their href, and I haven't been able to narrow down to the one that may be causing the issue, but I assume that since the edits are similar to the other pages, this would work


Roesh:
Hm, it looks like the deployment failed, and I get a 404 when I try to get details about why the failure occurred at https://vercel.com/blitzjs/blitzjs-com/AFtkqduZwPhgW6cLop7vpXb8fJcs

Not sure if I'm missing a step when creating that doc page

JuanM04:
@Roesh looks good!
If you try to build the page on your PC, you should be able to see the error. Nonetheless, I took a screenshot:
image

Roesh:
> ## Extending or replacing the default Blitz ESLint config {#extending}

Sounds good! I'll edit that, but per my first reply, I'm still not able to preview the build locally because of this error which occurs when I go the the documentation page, which I haven't been able to fix:
`The prop href expects a string or object in <Link>, but got undefined instead.`

flybayer:
Thank you!!

PR #479, opened 2021-05-18
Document authenticate.redirectTo: RouteManifest

Skn0tt:
Cc https://github.com/blitz-js/blitz/pull/2359


JuanM04:
Perfect. I think we should merge this to `canary`

PR #483, opened 2021-05-19
Update render docs to invert order and first prisma migrate so when b…

fotomotera:
…uilding getStaticPaths/Props doesn't error

I had trouble otherwise while using staticPaths/Props

### DIFF
--- a/app/pages/docs/deploy-render.mdx
+++ b/app/pages/docs/deploy-render.mdx
@@ -45,8 +45,8 @@ services:
     plan: starter
     buildCommand:
       yarn --frozen-lockfile --prod=false &&
-      blitz build &&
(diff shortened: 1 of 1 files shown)


app/pages/docs/deploy-render.mdx
```diff
@@ -45,8 +45,8 @@ services:
     plan: starter
     buildCommand:
       yarn --frozen-lockfile --prod=false &&
-      blitz build &&
-      blitz prisma migrate deploy
+      blitz prisma migrate deploy &&
+      blitz build
```
flybayer, line 49: `migrate deploy` should be after the build so the db isn't migrated in case build fails. But to fix the issue you can add `blitz prisma migrate` before `blitz build`

Issue #437, opened 2021-04-12
Light mode setting does not take effect on page load


flybayer:
1. Set light mode on the docs
1. Open blitzjs.com in new tab
1. See that dark mode is shown, even thought dark mode toggle is set to light mode

This used to work, so not sure why it's not now. We use https://github.com/pacocoursey/next-themes, so maybe upgrading that is enough? Or maybe we need to adjust our settings with how we use that.

JuanM04:
It seems like ThemeProvider doesn't update. Weird

JuanM04:
May be related to this: https://github.com/pacocoursey/next-themes/issues/40

5minpause:
I just tried it and couldn't reproduce the behaviour mentioned by @flybayer in https://github.com/blitz-js/blitzjs.com/issues/437#issue-856076141.
I believe #485 fixed it and this issue can be closed, @JuanM04 .

PR #485, opened 2021-05-21
Light/Dark mode fix

enemycnt:
Hello Blitz team, 
fix of #437 
I assume that dark mode is used by default, and light mode after switching.
Tested on chrome, firefox, and safari.


JuanM04 (approved):
Awesome, thanks!

PR #490, opened 2021-05-27
Add bilaterial reference between dynamic routing and useParam/useRouterQuery

antonykamp:
hey

I added some references from dynamic routing to useParam, useParams, useRemoteQuery, and reversed. 

antony


antonykamp:
Added package.lock by mistake

PR #493, opened 2021-05-29
Test database notes

ormarek:
- small improvements to docs (as discussed on discord) clarifying that test database must be different than main one.
- extension to `docker-compose.yml` containing pre-set test database - it has no volume on purpose as it's not something we need on test database


JuanM04 (approved):
Thank you!

PR #501, opened 2021-06-11
Clarify distinction between useSession & useCurrentUser

dbachrach:
Rereading through the docs it's unclear what the use-case difference is between `useSession()` and `useCurrentUser()`.

This removes the mentioning of suppressing the first render since it's unrelated to the useSession vs useCurrentUser discussion. 

I still think we need a *reason* that you would use useCurrentUser. I'm not really sure what that would be though?

### DIFF
--- a/app/pages/docs/authorization.mdx
+++ b/app/pages/docs/authorization.mdx
@@ -221,11 +221,6 @@ This is available on the client without making a network call to the
 backend, so it's available faster than the `useCurrentUser()` approach
 described below.
 
-However, due to the nature of static pre-rendering, the `session` will not
-exist on the very first render on the client. This causes a quick "flash"
-on first load. You can fix that by setting
-[`Page.suppressFirstRenderFlicker = true`](./pages##automatic-static-optimization)
-
 ```tsx
 import { useSession } from "blitz"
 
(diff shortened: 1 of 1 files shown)


JuanM04:
Awesome

> This removes the mentioning of suppressing the first render since it's unrelated to the useSession vs useCurrentUser discussion.

Yes, it's unrelated, but maybe it's worth mentioning it.

> I still think we need a reason that you would use useCurrentUser. I'm not really sure what that would be though?

Not all the data would be in `publicData`, since it would be a security issue, for example.

app/pages/docs/authorization.mdx
```diff
@@ -221,11 +221,6 @@ This is available on the client without making a network call to the
 backend, so it's available faster than the `useCurrentUser()` approach
 described below.
 
-However, due to the nature of static pre-rendering, the `session` will not
-exist on the very first render on the client. This causes a quick "flash"
-on first load. You can fix that by setting
-[`Page.suppressFirstRenderFlicker = true`](./pages##automatic-static-optimization)
-
```
flybayer, line 228: Yeah let's leave this as it's something people face

dbachrach:
Ok I put the sentence back but changed it from `However` to `Note` to better indicate that this is not in relation to the discussion of `useSession()` vs `useCurrentUser()`. I also made a note about using `useCurrentUser()` if you need access to secure data that cannot be placed in `publicData`.

JuanM04 (approved):
Nice, thanks!

PR #507, opened 2021-06-18
Add instruction for render.yaml

igeligel:
When working with TypeScript the build fails when we are not creating the model first. I can add a comment if you want mentioning that the models should be generated before the build first.

Here is an error from my recent project: Build error without generating models


flybayer:
Thanks!!

PR #518, opened 2021-07-10
Add info to set PORT for railway deployment

anothernode:
I'm just getting started with Blitz. Following the Getting Started Guide and the section about deployment on Railway I ran into an HTTP 502 Bad Gateway error.

I figured out that the port the Blitz server is running on must be configured manually for Railway. One way is to modify the start script as already [described in a discussion](https://github.com/blitz-js/blitz/discussions/2335). Another way is to add a `PORT` variable to the railway environment.

Maybe it would be a good idea to add this piece of information to the Railway Deployment section?

### DIFF
--- a/app/pages/docs/deploy-railway.mdx
+++ b/app/pages/docs/deploy-railway.mdx
@@ -57,10 +57,16 @@ The project ID is taken from the **Project Setup** page.
 
 4. Enter your required environment variables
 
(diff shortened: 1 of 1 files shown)


abuuzayr:
Absolutely right, @anothernode . Thank you for this PR!

abuuzayr (approved):
LGTM!

PR #520, opened 2021-07-11
`.nonstrict()` is deprecated and no longer needed

msichterman:


JuanM04 (approved):
I was so confused reading the changes :laughing:. Thank you!

PR #524, opened 2021-07-22
Fix unexpected inline code formatting in the tutorial

doi-t:
I've found that the inline code of `blitz generate` command is unexpectedly separated. This PR change them to a single inline code.

I've also updated the outputs with the latest texts.


abuuzayr:
great eye for detail, thank you @doi-t !

abuuzayr (approved):

PR #531, opened 2021-08-01
Update multitenancy.mdx

Talor-A:
fix mistake that causes `Role` to evaluate to `never[]`, see https://discord.com/channels/802917734999523368/871393674129461259/871490979646021642


flybayer:
Thank you!!

PR #533, opened 2021-08-04
Update `signup` mutation in multitenancy.mdx

m5r:
Fix the `signup` mutation to make the user insertion return a user's memberships.

See https://discord.com/channels/802917734999523368/872490237719371826/872490241469083688


flybayer:
Thanks!!

Issue #540, opened 2021-08-18
Dead link in session managment docs.


damilolarandolph:
The link in the "Customize Session Persistence & Database Access" section of session management docs is dead. It is meant to be a reference to the default session config that works with Prisma but leads to 404 on GitHub.

PR #541, opened 2021-08-18
Fix: dead link in session management docs

damilolarandolph:
Resolves: #540 

Fixed dead link in 'Customize Session Persistence & Database Access' section of the session management docs. Replaced the dead link with the current default implementation of `SessionConfig`.


flybayer:
thank you!

Issue #514, opened 2021-07-06
Is single threading a tradeoff of NextJS or NodeJS


jackbravo:
On https://github.com/blitz-js/blitzjs.com/blob/main/app/pages/docs/tradeoffs.mdx its stated that a single threaded server is:

> a tradeoff of Next.js specifically

But I've read that this tradeoff is generic to NodeJS event loop model. Which is which?

This doesn't mean that spawning background processes isn't still valid.

flybayer:
@jackbravo yes, good point! It is a Node thing. Want to PR a fix for that?

JuanM04:
Closed by #542

PR #542, opened 2021-08-22
Fix: clarify the single-threading tradeoff (#514)

mwcampbell:
I identified this tradeoff as belonging to Node.js generally rather than Next.js specifically. I also removed the bit about serverless from the heading, as it implied (at least to me) that the serverless deployment was somehow multithreaded. Instead, I added another paragraph explaining why this tradeoff is irrelevant on at least some serverless platforms, while noting that deployment to Vercel specifically is not recommended.


JuanM04 (approved):
Amazing, thanks

PR #544, opened 2021-08-28
Update deploy-railway.mdx

maciejmyslinski:
When deploying to railway without running migrations, no tables are set up in the database and the app can't work properly.


flybayer:
Thanks!

PR #555, opened 2021-09-28
add docs on client vs server

beerose:
I added a new page to the docs as per this comment: https://discord.com/channels/802917734999523368/891631382877569054/892170181458149416.

I wanted to summarize what's executed on the server versus what's on the client as it seems helpful to have this kind of summary. Let me know what you think!

I also wanted to include an architecture diagram but wasn't happy with how it looks and didn't want to modify the original svg too much. 

Though I think this diagram would be super helpful, so I'd create an issue for that after this PR is merged. Unless you like how it could look right now (it's not included in this PR):

Screenshot 2021-09-28 at 16 35 51

Screenshot 2021-09-28 at 16 35 45

### DIFF
--- a/app/pages/docs/client-and-server.mdx
+++ b/app/pages/docs/client-and-server.mdx
@@ -23,11 +23,8 @@ about [API Routes](./api-routes).
 
 #### React Components
 
-Everything inside a page or component file happens both server and client.
-However, you can opt-in for a full server-side rendering by exporting an
-async function called `getServerSideProps`. The server will call this
-function on every request. Read more about
-[`getServerSideProps`](./get-server-side-props).
+All code inside a page or component file runs on both server and client, except for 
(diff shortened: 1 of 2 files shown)


app/pages/docs/client-and-server.mdx
```diff
@@ -0,0 +1,38 @@
...
+API handlers defined in `api/` directory run only on the server. Read more
+about [API Routes](./api-routes).
+
+#### React Components
+
+Everything inside a page or component file happens both server and client.
+However, you can opt-in for a full server-side rendering by exporting an
+async function called `getServerSideProps`. The server will call this
+function on every request. Read more about
+[`getServerSideProps`](./get-server-side-props).
```
flybayer, line 30: Awesome, very good! Just one modification.

```suggestion
All code inside a page or component file runs on both server and client, except for 
[`getServerSideProps`](./get-server-side-props) or [`getStaticProps`](./get-static-props). Code that is imported and ran inside those two functions will only run on the server.
```

PR #557, opened 2021-09-29
updated command docker compose up -d

davyengone:
the `docker-compose up` command -d requires that the `docker-compose` package is installed on the user's computer in addition to the docker package.
Now, docker supports the `compose` option natively. You can use `docker compose` instead.


JuanM04 (approved):
Thank you!

PR #559, opened 2021-09-30
Update useQuery and useMutation docs

### LINKED ISSUE #548: Update useQuery docs with info on useErrorBoundary option
Hello,

I noticed that the [useQuery documentation](https://blitzjs.com/docs/use-query#options) includes few details about the `useErrorBoundary` option. I thought this could be confusing for someone unfamiliar with React Query. 

Specifically, the option is included in the function signature, but there is no information about it below. 

The [useMutation docs](https://blitzjs.com/docs/use-mutation#options) include plenty of information to get started using the option such as its default value. To my knowledge, this option works the same for both queries and mutations.

ajwgeek:
Updated the useQuery documentation to include a description of the useErrorBoundary option. I also tweaked the useMutation docs to include useErrorBoundary's type.

Resolves #548


JuanM04 (approved):
Thank you!

PR #582, opened 2021-10-20
docs(multitenancy): fix example update mutation

xkrsz:
In the current example, the code snippet is invalid. `update()` is limited to unique model fields, so the only way to execute this call is to add a unique constraint on `id` and `organizationId`. There's also `updateMany()`, but I wouldn't use that as an example as it will scan all table rows and can hurt performance. To my knowledge Prisma doesn't yet support limiting `updateMany()` to the first updated record.

Do let me know if explanation can be improved and if we should add more examples/info here.


flybayer:
Ah yes, I meant `findFirst` 🤦‍♂️

PR #593, opened 2021-11-02
Update --parent documenation in cli-generate.mdx

thisdotrob:
It took me longer than it should have done to understand the distinction between using the `--parent` flag and using `belongsTo:model` notation.

To generate the pages, mutations, queries AND model relationships for a parent:child pairing of models, you need to use both the `--parent` flag and set the `belongsTo:model` relationship.

I think the lines I have added to the docs are the minimal needed change to surface this info.


flybayer:
thank you!

PR #597, opened 2021-11-07
Update blitz config docs and add info about log type

lovethebomb:
Linked to https://github.com/blitz-js/blitz/pull/2888

### DIFF
--- a/app/pages/docs/blitz-config.mdx
+++ b/app/pages/docs/blitz-config.mdx
@@ -121,9 +121,8 @@ module.exports = {
 
 ### Log Type {#log-type}
 
-It is also possible to adjust the log type provided by Blitz, which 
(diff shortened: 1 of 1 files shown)


app/pages/docs/blitz-config.mdx
```diff
@@ -117,6 +119,23 @@ module.exports = {
 }
 ```
 
+### Log Type {#log-type}
+
+It is also possible to adjust the log type provided by Blitz, which 
+could be useful if you prefer to forward the log in a more usable
+format like JSON.
```
beerose, line 126:
```suggestion
It is also possible to adjust the logging type, which could 
be helpful if you prefer to forward the logs in JSON format.
```

beerose (approved):

PR #599, opened 2021-11-10
Improve the order of steps in tutorial

### LINKED ISSUE #594: Move mention of `blitz console` after updating generated code in the tutorial
We should probably move `Playing with the Prisma database client` after the `Update generated code for our model attributes` in the tutorial, as `blitz console` fails before updating the generated code.

Related discussion: https://discord.com/channels/802917734999523368/870041934046060615/905449030073602128

5minpause:
The previous order asked you to run `blitz console` earlier. That command resulted in errors since the generated files had to be updated first.
The unused file `app/choices/mutations/createChoice.ts` also needs to be deleted earlier.

Fixes #594

### DIFF
--- a/app/pages/docs/tutorial.mdx
+++ b/app/pages/docs/tutorial.mdx
@@ -551,8 +551,9 @@ export default resolver.pipe(
 This mutation will now delete the choices associated with the question
 prior to deleting the question itself.
 
-## Cleanup {#cleanup}
+## Remove unnecessary file {#remove-unnecessary-file}
 
+Our scaffolding created a mutation file for us that is no longer needed.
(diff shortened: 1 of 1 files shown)


app/pages/docs/tutorial.mdx
```diff
@@ -606,6 +551,67 @@ export default resolver.pipe(
 This mutation will now delete the choices associated with the question
 prior to deleting the question itself.
 
+## Cleanup {#cleanup}
```
beerose, line 554: Maybe we can call it something like "Remove unnecessary file"? (As "cleanup" sounds better if it's in the end — not strong opinions, though)

5minpause: Sounds like a good plan. On it.

PR #601, opened 2021-11-11
fix: typo on folder name - Session Management page

danestves:
This is a simple PR to replace `muations` with `mutations`


beerose (approved):
Thanks!

PR #615, opened 2021-11-27
Add the build script within code block

ajanth97:
This is a small change to include the build script within a code block in line 81 and line 82


JuanM04 (approved):
Nice!

PR #618, opened 2021-11-30
docs: add invokeWithMiddleware for server queries

david-arteaga:
Use `invokeWithMiddleware` when calling queries on the server in `getServerSideProps`


JuanM04 (approved):
Awesome, thanks

Issue #606, opened 2021-11-17
`createQuestionSchema` seems not working (tutorial)


exKAZUu:
### What is the problem?

The tutorial shows the following code, but client validation seems not working.

```
// app/pages/questions/validations.ts
import { CreateQuestion } from "app/questions/mutations/createQuestion"

export const createQuestionSchema = CreateQuestion
```

For example, if we change the definition of `CreateQuestion` as follows:

```
// app/questions/mutations/createQuestion.ts
export const CreateQuestion = z.object({
  text: z.string().email(),
  choices: z.array(z.object({text: z.string()})),
})
```

I expect we cannot enter a non-email string in a question text, but we can do and we will get the following error from a SERVER (not a client).

![image](https://user-images.githubusercontent.com/436237/142145404-5b5d71a3-2ded-415c-b3c5-c9e6163d1a09.png)

If we move the definition of `CreateQuestion` from `app/questions/mutations/createQuestion.ts` to `app/pages/questions/validations.ts` as follows, then client validation works well.

```
// app/pages/questions/validations.ts
export const CreateQuestion = z.object({
  text: z.string().email(),
  choices: z.array(z.object({text: z.string()})),
})
```

```
// app/questions/mutations/createQuestion.ts
import { CreateQuestion } from "../../pages/questions/validations";
```

### Paste all your error logs here:

Please see the above screenshot.

### Paste all relevant code snippets here:

Please see the above code snippets.

### What are detailed steps to reproduce this?

1. Follow the tutorial (https://blitzjs.com/docs/tutorial)
2. Change the definition of `CreateQuestion` (`text: z.string()` -> `text: z.string().email()`)
3. Enter a non-email string in the question text
4. Client validation doesn't be triggered (SHOULD BE TRIGGERED)

### Run `blitz -v` and paste the output here:

```
macOS Monterey | darwin-x64 | Node: v16.11.1

blitz: 0.42.4 (local)

  Package manager: yarn
  System:
    OS: macOS 12.0.1
    CPU: (8) x64 Intel(R) Core(TM) i7-1068NG7 CPU @ 2.30GHz
    Memory: 26.34 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.11.1 - /private/var/folders/tj/9535q_6j1h55lgqwzlgg1mwc0000gn/T/xfs-5e9be5e5/node
[...]

PR #619, opened 2021-12-02
docs: move CreateQuestion declaration in validations.ts

exKAZUu:
follow up for #606 and 38044d25dfca3704d705532b35d270980ccc2bfa

This PR is for the following issue described in #606.

> If we move the definition of CreateQuestion from app/questions/mutations/createQuestion.ts to app/pages/questions/validations.ts as follows, then client validation works well.

### DIFF
--- a/app/pages/docs/tutorial.mdx
+++ b/app/pages/docs/tutorial.mdx
@@ -645,21 +645,46 @@ export function QuestionForm<S extends z.ZodType<any, any>>(
 }
 ```
 
-Now open `app/questions/mutations/createQuestion.ts` and update the zod
-schema so that the choice data is accepted in the mutation. We also need
-to update the `db.question.create()` call so that the choices will be
-created. After that we need to export the `CreateQuestion` zod schema
-because we will be using it in the next step to create a validation schema
-for our `QuestionForm`.
+Next we're going to create a separate file to store the validation schema
+for our `QuestionForm`. In the `app/questions` folder create a new file
+called `validations.ts` and copy `CreateQuestion` declaration from
+`./mutations/createQuestion.ts`. Then update the zod schema so that
+the choice data is accepted in the mutation., and export it so
+we can use it for our `QuestionForm` schema in the later step.
+
+```diff
+// app/questions/validations.ts
++ import {z} from 'zod'
+
++ export const CreateQuestion = z
++    .object({
++      text: z.string().nonempty({message: "You must enter a question."}),
++     choices: z.array(z.object({text: z.string()})),
++    })
+```
+
+<Card type="info">
+  We create a shared `validations.ts` file because we cannot import
+  anything from a query (or mutation) file other than the query itself
+  into the client. You can read more about why in [Query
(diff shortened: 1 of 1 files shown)


app/pages/docs/tutorial.mdx
```diff
@@ -645,21 +645,46 @@ export function QuestionForm<S extends z.ZodType<any, any>>(
...
-    })
++    })
+```
+
+<Card type="info">
+  We create a shared `validations.ts` file because we cannot import
+  anything from a query (or mutation) file other than the query itself
+  into the client. You can read more about why in [Query
+  Usage](https://blitzjs.com/docs/query-usage) and [Mutation
+  Usage](https://blitzjs.com/docs/mutation-usage).
```
beerose, line 671: Simplified it a bit, but feel free to disagree 😅
```suggestion
  We create a shared `validations.ts` file so that we can import
  the schemas and use them on the client. We're doing it because
  the code in the `queries/` and `mutations` folders runs only on 
  the server.
```

app/pages/docs/tutorial.mdx
```diff
@@ -645,21 +645,46 @@ export function QuestionForm<S extends z.ZodType<any, any>>(
...
-schema so that the choice data is accepted in the mutation. We also need
-to update the `db.question.create()` call so that the choices will be
-created. After that we need to export the `CreateQuestion` zod schema
-because we will be using it in the next step to create a validation schema
-for our `QuestionForm`.
+Next we're going to create a separate file to store the validation schema
+for our `QuestionForm`. In the `app/questions` folder create a new file
+called `validations.ts` and copy `CreateQuestion` declaration from
+`./mutations/createQuestion.ts`. Then update the zod schema so that
+the choice data is accepted in the mutation., and export it so
```
beerose, line 652:
```suggestion
the choice data is accepted in the mutation, and export it so
```

beerose, line 652:
```suggestion
`./mutations/createQuestion.ts`. Then update the zod schema to
validate the choices array, and export it so
```

app/pages/docs/tutorial.mdx
```diff
@@ -645,21 +645,46 @@ export function QuestionForm<S extends z.ZodType<any, any>>(
...
+  We create a shared `validations.ts` file because we cannot import
+  anything from a query (or mutation) file other than the query itself
+  into the client. You can read more about why in [Query
+  Usage](https://blitzjs.com/docs/query-usage) and [Mutation
+  Usage](https://blitzjs.com/docs/mutation-usage).
+</Card>
+
+Now open `app/questions/mutations/createQuestion.ts`, remove old
+CreateQuestion declaration and import new CreateQuestion from
+../validations.ts. We also need to update the `db.question.create()`
```
beerose, line 676: Also, maybe we can remove it while copying it to a separate file?
```suggestion
Now open `app/questions/mutations/createQuestion.ts`, remove old
`CreateQuestion` declaration and import new `CreateQuestion` from
`../validations.ts`. We also need to update the `db.question.create()`
```

Issue #588, opened 2021-10-25
Add docs on persistent layouts and layouts in general


beerose:
Related discussion: https://github.com/blitz-js/blitz/discussions/2866#discussioncomment-1517018

PR #625, opened 2021-12-09
Remove reference to unexisting page

SofianeDjellouli:


beerose (commented):
Not sure why the page doesn't exist (maybe it was removed?), but we probably could add a layout page with this content: https://nextjs.org/docs/basic-features/layouts?

beerose:
Let's merge this, and here's an issue for the missing layouts docs: #588

beerose (approved):

PR #629, opened 2021-12-13
Fix import of log in custom server example

zenhob:
The import of `log` in the docs example no longer works in the latest version of Blitz.


beerose (approved):
Thanks! We also have an issue here to expose it from `blitz`: https://github.com/blitz-js/blitz/issues/2782.

PR #631, opened 2021-12-15
Small doc correction and code snippets update (new Routes object)

kreako:
I just read a part of the documentation and found :

- a small error "from" -> "form"
- a small error on undefined "organizationId" binding 
- outdated code snippets on tutorial (using plain strings instead of Routes object)

Thanks for your amazing work !


beerose (approved):
Thanks for your PR!!

PR #643, opened 2021-12-29
Update showcaseList.js

felixyeboah:


JuanM04 (changes requested):
Amazing! Please, add the thumbnail image to `public/showcase`

felixyeboah:
I do not have rights. I tried

beerose:
> I do not have rights. I tried

Hi @felixyeboah! What Juan meant is to add an image to the public directory, and then commit and push. You already created a PR so looks like it's working fine for you. What do you mean by not having rights?

PR #645, opened 2021-12-31
fix HTML titles are now displayed in HTML tag format.

### LINKED ISSUE #644: dog page Searcher Modal title bug
_ErrorBoundary__Component_-_Blitz_js_-_Vivaldi-2
Probably "&gt;" is displayed where "<" or ">" should be displayed.

mochi-sann:
fix #644 
## before 
image
## after
Future_of_Blitz_-_Vivaldi
r

### DIFF
--- a/app/core/components/Search.js
+++ b/app/core/components/Search.js
@@ -38,10 +38,6 @@ export function Search({className = ""}) {
     [setIsOpen, setInitialQuery],
   )
 
-  const convertToHtmlTag = (value) => {
-    return value.replace(/&amp;/g, "&").replace(/&lt;/g, "<").replace(/&gt;/g, ">")
-  }
-
   useDocSearchKeyboardEvents({
     isOpen,
     onOpen,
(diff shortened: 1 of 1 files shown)


app/core/components/Search.js
```diff
@@ -37,6 +37,9 @@ export function Search({className = ""}) {
     },
     [setIsOpen, setInitialQuery],
   )
+  const ConvertToHtmlTag = (value) => {
+    return value.replace(/&amp;/g, "&").replace(/&lt;/g, "<").replace(/&gt;/g, ">")
+  }
```
beerose, line 42: Let's keep the function name lowercase
```suggestion

  const convertToHtmlTag = (value) => {
    return value.replace(/&amp;/g, "&").replace(/&lt;/g, "<").replace(/&gt;/g, ">")
  }
```

beerose (commented):
Thanks!

mochi-sann:
Thank you for your review!

JuanM04 (changes requested):
Nice!
I noticed that `hierarchy.lvl0` doesn't need to be converted anymore, so the code can be reduced to this:

```js
// The titles are encoded, so we need to decode them.
// &lt;Script&gt; --> <Script>
const _highlightResult = {...item._highlightResult}

_highlightResult.hierarchy.lvl0.value = _highlightResult.hierarchy.lvl0.value
  .replace(/&amp;/g, "&")
  .replace(/&lt;/g, "<")
  .replace(/&gt;/g, ">")

return {
  ...item,
  _highlightResult,
  url: `${a.pathname}${hash}`,
}
```

mochi-sann:
Thank you ,review
I've reduced the code as you pointed out.

PR #649, opened 2022-01-17
document logger factory config

nakleiderer:
See also: https://github.com/blitz-js/blitz/pull/3131


nakleiderer:
Closing because https://github.com/blitz-js/blitz/pull/3131 has been closed and this documentation is not needed.

PR #670, opened 2022-03-28
updating supertokens url in auth docs

jscyo:
SuperTokens recently changed its URL from supertokens.io -> supertokens.com, this PR reflects that change


beerose (approved):
Thanks!

PR #671, opened 2022-04-01
update faker.js reference

Caslus:
marak's faker.js has been gone for a while now, I've updated the database-seeds page to refer to a new faker alternative.


beerose (approved):
Thanks!

PR #677, opened 2022-04-29
Update auth-utils.mdx

the-bayer:
Noticed this situation when creating an API key

### DIFF
--- a/app/pages/docs/auth-utils.mdx
+++ b/app/pages/docs/auth-utils.mdx
@@ -125,6 +125,8 @@ This is a convenience wrapper that uses the node
 the `sha256` algorithm. It is used for things like hashing password reset
 tokens before saving them in the database.
 
+Hash256 is also useful for storing strings like API keys in the database because the returned hash will always be the same for a given string. Therefore, you can still verify that an API key exists in the database when the only value you have to reference is the hashed key.
+
 #### Example Usage
 
 ```ts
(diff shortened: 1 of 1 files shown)


app/pages/docs/auth-utils.mdx
```diff
@@ -153,6 +155,8 @@ This is used when a user sets a new password.
 It takes a password string and returns a secure hash for storing in your
 database.
 
+SecurePassword.hash will return a different hash when given the same string, hence the necessity of SecurePassword.verify to compare hashs.
```
Kosai106, line 158:
```suggestion
`SecurePassword.hash` will return a different hash when given the same string, hence the necessity of `SecurePassword.verify` to compare hashes.
```

app/pages/docs/auth-utils.mdx
```diff
@@ -125,6 +125,8 @@ This is a convenience wrapper that uses the node
 the `sha256` algorithm. It is used for things like hashing password reset
 tokens before saving them in the database.
 
+Hash256 is also useful for storing strings like API keys in the database because the returned hash will always be the same for a given string. Therefore, you can still verify that an API key exists in the database when the only value you have to reference is the hashed key.
```
Kosai106, line 128:
```suggestion
`sha256` is also useful for storing strings like API keys in the database because the returned hash will always be the same for a given string. Therefore, you can still verify that an API key exists in the database when the only value you have to reference is the hashed key.
```

PR #710, opened 2022-07-22
fix dead link

iojcde:
The named anchor `https://github.com/blitz-js/blitz#sponsors-and-donations` doesn't exist anymore.
Replaced it with `https://github.com/blitz-js/blitz#financial-contributors`.


beerose (approved):
Thanks!

PR #711, opened 2022-07-23
docs: Wrap passport example with api() from blitz-server

iojcde:
No clue why #700 was closed, this is exactly the same change.

`PassportAuth()` should be wrapped with `api()`, similar to how it's done at https://github.com/blitz-js/blitzjs.com/blob/d61b7176f9d5789397d0b8954cd64eb25bd26077/app/pages/docs/rpc-setup.mdx#blitz-rpc-with-nextjs-setup-nextjs.

Fixes blitz-js/blitz#3467.

### DIFF
--- a/app/pages/docs/passportjs.mdx
+++ b/app/pages/docs/passportjs.mdx
@@ -20,9 +20,10 @@ contents.
 ```ts
 // app/api/auth/[...auth].ts
 import { passportAuth } from "blitz"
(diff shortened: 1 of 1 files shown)


beerose (approved):
Thanks!

googamanga:
@beerose How would this be handled on v 0.45.x?

PR #728, opened 2022-08-12
Fix imports and wrappers in the docs

siddhsuresh:
- [x] Fix missing gSP wrapper for getStaticProps in https://blitzjs.com/docs/query-usage#get-static-props
- [x] Fix imports in paginated query page example - https://blitzjs.com/docs/use-paginated-query#example


beerose (approved):
Thank you 🙏🙏🙏

PR #731, opened 2022-08-13
Update installation command

makeitglitch:
Don't suggest installing `alpha` - to match https://github.com/blitz-js/blitz/blob/main/README.md#install-blitz


beerose (approved):
Thank you!

PR #737, opened 2022-08-16
Add product hire.win to showcase

vivek7405:
- Added my product hire.win to showcase as suggested by @beerose


beerose (approved):
Awesome, thank you!

PR #743, opened 2022-08-19
Update eslint-config.mdx to the Blitz Tooklit configuration

siddhsuresh:
- [x] change blitz.config.js to next.config.js
- [x] change ```module.exports = { extends: ["blitz"]}```  to ```module.exports = require("@blitzjs/next/eslint")```
- [x] changing extends from ```blitz``` to ```require.resolve('@blitzjs/next/eslint')```


FlorianWendelborn (approved):

app/pages/docs/eslint-config.mdx
```diff
@@ -34,7 +32,7 @@ In the below example:
 
 ```js
 module.exports = {
-  extends: ["blitz", "shared-config"],
+  extends: ["./node_modules/@blitzjs/next/eslint", "shared-config"],
```
beerose, line 35: Hmm, doesn't `@blitzjs/next/eslint` work here?

beerose (commented):
Thanks a lot for the PR!! Left one question

app/pages/docs/eslint-config.mdx
siddhsuresh: > Hmm, doesn't `@blitzjs/next/eslint` work here?

As the .eslint config is present locally in the @blitzjs/next folder in node_modules, the eslint docs https://eslint.org/docs/latest/developer-guide/shareable-configs needs to either link the eslint file or reference it with the absolute path to the config that needs to be extended from.

image

@beerose Would this syntax be preferred?
```js
module.exports = {
    extends: require.resolve('@blitzjs/next/eslint-config'),
    rules: {
      "additional-rule": "warn",
    },
  }
```

beerose (approved):
Thanks!

PR #746, opened 2022-08-22
fix: unable to drag the scroll over the hand

iquirino:
With "pointer-events: none;" we are able to drag scrollbar even with the mouse over the hand


beerose (commented):
Thanks for the PR!

Could you describe a bit more what the issue was? I haven't noticed any problems with scrolling, and not sure where to look.

iquirino:
Just a pixel perfect work. 

I attached some screen recordings to show you:

Screen sample of blitzjs.com
blitzjs-default

Screen sample with `pointer-events: none`
blitzjs-pointer-events-none

https://css-tricks.com/almanac/properties/p/pointer-events/

beerose:
Oh okay! Thanks a lot for explaining!

beerose (approved):

PR #748, opened 2022-08-23
improved Image component document for Blitz 2.0

DataCrusade1999:
The previous `Image` component documentation said that In order to add domains for external images we have to edit `blitz.config.js` but in **Blitz 2.0** after running `blitz new project-name`  no file named `blitz.config.js` is created in the project.
the configuration has to be made in `next.config.js` This is what I added to the documentation along with an example configuration where I showed the process.


app/pages/docs/image-component.mdx
```diff
@@ -94,7 +94,7 @@ and `blurDataURL` manually.
 
 In addition to using properties available to the `Image` component, you
 can optionally configure Image Optimization for more advanced use cases
-via `blitz.config.js`.
+via `blitz.config.js`. Use `next.config.js` if you are using Blitz 2.0.
```
beerose, line 97: Since this documentation is only for Blitz 2.0, we can just say:
 ```suggestion
via `next.config.js`.
```

**Review by beerose — COMMENTED** [review]
Thanks a lot for taking the time to fix this!

One thing: let's only mention `next.config.js`. We don't have to include info about `blitz.config` in the current docs.

File: app/pages/docs/image-component.mdx (hunk 1 above)
Comment (DataCrusade1999) [reply app/pages/docs/image-component.mdx]: I've made the changes that you suggested `blitz.config.js` is not mentioned now and I've also removed the previous configuration example because it was an example for `blitz.config.js`.

File: app/pages/docs/image-component.mdx [hunk 2]
```diff
@@ -103,12 +103,23 @@ an absolute url for the Image `src` and specify which `domains` are
 allowed to be optimized. This is needed to ensure that external urls can't
 be abused.
 
+Starting from Blitz 2.0 you need to make some configuration changes
+in `next.config.js` In order to add external domains into your project.
```
beerose, line 107:
```suggestion
To add external domains to your project, you need to add the `images` property to your `next.config.js` file: 
```

app/pages/docs/image-component.mdx
```diff
@@ -103,12 +103,23 @@ an absolute url for the Image `src` and specify which `domains` are
...
+// @ts-check
+const { withBlitz } = require("@blitzjs/next")
+
+/**
+ * @type {import('@blitzjs/next').BlitzConfig}
+ **/
+const config = {
   images: {
-    domains: ["example.com"],
+    domains: ["www.example.com"],
```
beerose, line 118:
```suggestion
    domains: ["www.example.com"], // hightlight-line
```

PR #750, opened 2022-08-23
Improve the @blitzjs/next documentation

dillonraphael:
- Add __Example__ headings to client & server sections of @blitzjs/next
- Add __API__ headings with __Arguments__ & __Returns__ for both the client & server sections of @blitzjs/next

### DIFF
--- a/app/pages/docs/blitzjs-next.mdx
+++ b/app/pages/docs/blitzjs-next.mdx
@@ -19,7 +19,8 @@ npm i @blitzjs/next # yarn add @blitzjs/next # pnpm add @blitzjs/next
 ## Client {#blitzjs-next-client}
 
 ### Example {#blitzjs-next-client-example}
-Inside `app > blitz-client.ts`
+
+Inside `app/blitz-client.ts`:
 
 ```ts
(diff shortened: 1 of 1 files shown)


app/pages/docs/blitzjs-next.mdx
```diff
@@ -18,6 +18,7 @@ npm i @blitzjs/next # yarn add @blitzjs/next # pnpm add @blitzjs/next
 
 ## Client {#blitzjs-next-client}
 
+### Example {#blitzjs-next-client-example}
 Inside `app > blitz-client.ts`
```
beerose, line 22:
```suggestion
### Example {#blitzjs-next-client-example}

Inside `app/blitz-client.ts`:
```

beerose: I would remove the `Example` heading and tell users that they must create a `blitz-client.ts` file in <root>/app directory (or somewhere else). Same for the server configuration. In other words — I would write these docs with more explanations on why and what users are supposed to do.

beerose (changes requested):

flybayer (approved):

beerose (approved):

PR #755, opened 2022-09-02
Add back recipe docs

dillonraphael:
Add back the recipe documentation.

Todo:

- [ ] Update `Modifying Blitz config -> Modifying Next Config`
- [ ] Adding a middleware to `blitz-server.ts`


beerose (changes requested):
Need to also add `cli-install` to the CLI section in the docs.

PR #757, opened 2022-09-08
Fix Showcase page

StillScripts:
Combox is an inactive domain and shouldn't be displayed on the Showcase page. 
[View Website](https://combox.app/)


beerose (approved):
Thanks!

PR #770, opened 2022-09-22
Set the server port as provided by Render

nerixim:
Hello and thanks for your work.
It's my first time using Render and I used your template, but the builds were failing until I set the server port.
I believe it would be a better developer experience if you could copy the blueprint and it just worked 😇


beerose (approved):
Thanks!

PR #795, opened 2022-11-10
includeRPCFolder documentation for monorepo

paulm17:
New documentation for the PR: https://github.com/blitz-js/blitz/pull/3953


dillonraphael:
@paulm17 This is well written, but can be simpler in my opinion. Instead of creating a monorepo page, we should update the `rpc-specifications` page. 

Right under `Changing resolverPath option` you can create another subheader called `Resolvers in a separate folder`

dillonraphael (approved):

PR #811, opened 2023-03-24
Next 13 App Directory Docs

siddhsuresh:
Docs for https://github.com/blitz-js/blitz/pull/4079/ and https://github.com/blitz-js/blitz/pull/4103

### DIFF
--- a/app/pages/docs/usage-next-13.mdx
+++ b/app/pages/docs/usage-next-13.mdx
@@ -33,6 +33,29 @@ Add the new `use client` directive to the following files:
 
 This provider should wrap the app and should be placed at the `(root)/layout.ts` file.
 
+**Setup** 
+
+```ts
+// src/blitz-client
+
+"use client"
+import {AuthClientPlugin} from "@blitzjs/auth"
+import {setupBlitzClient} from "@blitzjs/next"
+import {BlitzRpcPlugin} from "@blitzjs/rpc"
+
+export const {withBlitz, BlitzProvider} = setupBlitzClient({
+  plugins: [
+    AuthClientPlugin({
+      cookiePrefix: "web-cookie-prefix",
+    }),
+    BlitzRpcPlugin({}),
+  ],
+})
+
+```
+
+**In root layout.(ts|js) file**
+
 ```tsx
 // layout.ts
 import BlitzProvider from "src/blitz-client"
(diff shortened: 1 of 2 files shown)


app/pages/docs/usage-next-13.mdx
```diff
@@ -0,0 +1,201 @@
...
+#### Codemods {#nextjs-codemods}
+
+https://nextjs.org/docs/advanced-features/codemods
+
+### Support with the App Router (Beta) {#app-directory}
+
+To support the new paradigm of Next that uses [React Server Side Components](https://github.com/reactjs/rfcs/blob/main/text/0188-server-components.md), 
+the following methods and hooks have been implemented to work in the new `app` directory
+
+#### Required changes {#required-changes}
```
flybayer, line 25: need to add that if they are using legacy blitz `app` setup, they need to rename that to `src`

app/pages/docs/usage-next-13.mdx
```diff
@@ -0,0 +1,201 @@
...
+import getCurrentUser from "src/users/queries/getCurrentUser"
+```
+
+Now, we can directly invoke the resolver using the invoke function
+
+```tsx
+const user = await invoke(getCurrentUser, null)
+```
+
+#### Compiled Changes to the Blitz-Server and Blitz-Client Files {#changes-required-blitz-files}
```
flybayer, line 159: what is this section?

app/pages/docs/usage-next-13.mdx
```diff
@@ -0,0 +1,201 @@
+---
+title: Usage Guide - Next.js 13 with Blitz Toolkit
+sidebar_label: Usage Guide Next.js 13
+---
+
+### Migration Guide {#migration-guide}
+
+Blitz **does not require any breaking changes** with the major upgrade of `Next 12` ➔ `Next 13` for usage with the `pages` directory.
```
flybayer, line 8:
```suggestion
Blitz **does not require any breaking changes** with the major upgrade of `Next 12` ➔ `Next 13` for usage with the `pages` directory. You can even continue to use the old blitz layout with all your files in `app`. But you'll need to rename that directory if you want to use the new Next.js app router.
```

flybayer (changes requested):

flybayer (approved):

PR #814, opened 2023-04-02
chore(docs): correct list order

Kosai106:
The list was out of order which caused some confusion (as evident by my initial commit).


siddhsuresh (approved):
Thanks for the PR! LGTM

PR #822, opened 2023-05-24
(blitz-auth-with-next.mdx) Make the Github link absolute to prevent appending it to the current path

nerixim:
Make the Github link absolute to prevent appending it to the current path.


siddhsuresh (approved):
Thank you for the PR!

PR #827, opened 2023-06-23
Docs: Routes: Section on hash URLs

tordans:
This will add a section on https://blitzjs.com/docs/route-manifest#query-parameters that describes how to use Hash URLs with Blitz and Next JS.

See also Discord.


siddhsuresh (approved):
LGTM

PR #830, opened 2023-08-03
sponser: add `Byteflow` as a bronze sponser

siddhsuresh:


app/core/components/SponsorPack.js
```diff
@@ -33,6 +33,13 @@ const sponsors = [
     tier: 3,
     cost: 250,
   },
+  {
+    name: "Byteflow",
+    href: "https://byteflow.dev/?ref=blitzjs",
```
siddhsuresh, line 38:
```suggestion
    href: "https://byteflow.app/?ref=blitzjs",
```

PR #843, opened 2023-10-10
Update usage-next-13.mdx: remove ".js" file ending

tordans:
My understanding is, that the current blitz version only generates .ts files anymore. We can clean this up a bit.


siddhsuresh (approved):
Thanks!

tordans:
This is also part of https://github.com/blitz-js/blitzjs.com/pull/844/files#diff-4749f563013949111052bf8f047307905982bb9fc173210fe6b244e448df2ad6R66 (by accident). I will close this PR since #844 is more important.

PR #844, opened 2023-10-10
Update usage-next-13.mdx: Still use `authConfig`

tordans:
This is how the auth docs show it (https://blitzjs.com/docs/blitz-auth-with-next#add-blitz-server) and I assume all blitz apps with auth are set up. It is confusing when the app router docs show it differently.

At least I assume that the blitz-server still needs to share the cookie name(?)


siddhsuresh (approved):
Thanks for the PR

tordans:
@siddhsuresh I updated the PR to follow your setup in https://github.com/blitz-js/blitz/pull/4146/files which has a separate file src/blitz-auth-config.ts. I had to apply this setup for our migration as well, because importing directly caused issues because I imported from a server component file into a client component (or was it the other way around…?). Using the patter you used in your PR solves this.

siddhsuresh (approved):

PR #845, opened 2023-10-10
Update usage-next-13.mdx: Clarify `/src/app` folder

tordans:
Update: I tested this again with my test app at https://github.com/FixMyBerlin/blitz-test and it turns out you can run the page in root `/app/*` and include from `/src/*`. 

I updated this PR to a recommendation, because I still think its better to use `/src/app` for all blitz apps that use `/src/pages`.


siddhsuresh (approved):
Thanks for the PR!

PR #846, opened 2023-10-10
Update usage-next-13.mdx: Hint at `loading.tsx` files

tordans:
The fact that layouts need to be handled differently in the app folder becomes very clear right from the start. However, the possible changes using the special `loading` files is a bit more hidden, which is why I suggest hinting at it in the migration docs.

### DIFF
--- a/app/pages/docs/usage-next-13.mdx
+++ b/app/pages/docs/usage-next-13.mdx
@@ -29,6 +29,12 @@ Add the new `use client` directive to the following files:
 1. `src/blitz-client.(ts|js)`
 2. All Files with usage of `useQuery`, `useInfiniteQuery`, `usePaginatedQuery`, `useMutation`, `Hydrate` and other React Query client side hooks.
 
+#### Migrate `<Suspense>` to `loading.tsx`
+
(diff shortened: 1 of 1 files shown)


siddhsuresh (commented):
So while technically true, a `loading.tsx` file will wrap the entire `page.tsx` inside a suspense boundary, which users might not expect. Unless the dev is using `invoke` to get the data and render in `page.tsx`. 

Otherwise keeping `<Suspense>` would be a good bet to suspend on the required parts of the UI especially when using `useQuery` and its siblings.

tordans:
@siddhsuresh looking at a regular pages-page like https://github.com/blitz-js/blitz/blob/main/packages/generator/templates/page/__modelIdParam__.tsx#L84-L86 everything except for the one Link is part of the Suspense. Which is IMO 99% of what `loading.tsx` does, is it not? The only think i loos it this one link which only shows up after loading is done. I gain, however, a much simpler `__modelIdParam__.tsx` file with just one regular component.

The `loading.tsx` is still wrapped inside the nested layouts that apply to it, so it's still a Suspense/loading-animation in the context of the layout — AKA this part does not change.

PR #847, opened 2023-10-13
Update usage-next-13.mdx: `BlitzProvider` as client component

tordans:
Theres is also the request for help at https://discord.com/channels/802917734999523368/1161576668863922176. 

See also https://github.com/blitz-js/blitz/issues/4232.

## Change

My understanding is, that `BlitzProvider` is a client component and the way I understand the app directory means that everything below in the include-tree needs to be a client component as well.

## Testcase

1. checkout https://github.com/FixMyBerlin/blitz-test/commits/blitzprovider-test which is the error case without the use client in the layout file that holds the blitzprovider
2. See error

**Error**
> Unhandled Runtime Error
Error: Unsupported Server Component type: undefined

image

### DIFF
--- a/app/pages/docs/usage-next-13.mdx
+++ b/app/pages/docs/usage-next-13.mdx
@@ -24,14 +24,13 @@ the following methods and hooks have been implemented to work in the new `app` d
 
 #### Required changes {#required-changes}
 
-Add the new `use client` directive to the following files:
+Add the new `use client` directive to **all** files in the include-tree below [`BlitzProvider`](https://blitzjs.com/docs/usage-next-13#blitz-provider), which is a client component.
 
-1. `src/blitz-client.(ts|js)`
-2. All Files with usage of `useQuery`, `useInfiniteQuery`, `usePaginatedQuery`, `useMutation`, `Hydrate` and other React Query client side hooks.
+This includes `src/blitz-client.(ts|js)` and all Files with usage of `useQuery`, `useInfiniteQuery`, `usePaginatedQuery`, `useMutation`, `Hydrate` and other React Query client side hooks or browser actions like `onClick` ([more in the NextJS Docs](https://nextjs.org/docs/app/building-your-application/rendering/client-components#using-client-components-in-nextjs)).
 
 #### BlitzProvider {#blitz-provider}
 
-This provider should wrap the app and should be placed at the `(root)/layout.ts` file.
+This client component is a provider that should wrap the app and should be placed at the `(root)/layout.ts` file. 
 
 **Setup** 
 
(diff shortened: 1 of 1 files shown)


siddhsuresh:
@tordans I do not understand this, `BlitzProvider` is a client component and takes the rest of the application as props. We are following this pattern https://nextjs.org/docs/app/building-your-application/rendering/composition-patterns#supported-pattern-passing-server-components-to-client-components-as-props

We can definitely use server components with blitz.

app/pages/docs/usage-next-13.mdx
```diff
@@ -58,6 +57,8 @@ export const {withBlitz, BlitzProvider} = setupBlitzClient({
 
 ```tsx
 // layout.ts
+'use client'
```
siddhsuresh, line 60: not required

app/pages/docs/usage-next-13.mdx
```diff
@@ -24,14 +24,13 @@ the following methods and hooks have been implemented to work in the new `app` d
 
 #### Required changes {#required-changes}
 
-Add the new `use client` directive to the following files:
+Add the new `use client` directive to **all** files in the include-tree below [`BlitzProvider`](https://blitzjs.com/docs/usage-next-13#blitz-provider), which is a client component.
```
siddhsuresh, line 27: `all` files?

siddhsuresh (changes requested):

app/pages/docs/usage-next-13.mdx
tordans: The thing is, that https://nextjs.org/docs/app/building-your-application/rendering/composition-patterns#unsupported-pattern-importing-server-components-into-client-components

> You cannot import a Server Component into a Client Component

Results is a bad error message 

> Error: Unsupported Server Component type: undefined

One thing I did was to remove all barrel files from our project in order to work around this issue, as suggested in https://github.com/vercel/next.js/issues/41940

(But in the end we added `'use client'` to all files (for now), just to get to a working version.)

My current understanding is: Since I cannot import server components into client components and I cannot audit every include, for [the slot-workaround](https://nextjs.org/docs/app/building-your-application/rendering/composition-patterns#supported-pattern-passing-server-components-to-client-components-as-props) you mentioned, I have to add `use client` to all files that are imported in other files.

app/pages/docs/usage-next-13.mdx
tordans: I am happy to learn how :-). The "Testcase" I described above shows the issue. Add/remove the "use client" and it works or doen't…

tordans:
The changes I propose here are wrong, I will close the PR.

